PR #20, opened 2022-09-22
feat(xml): support metasheet `@TABLEAU` and template option

huieric:
New features:
- support metasheet like `@TABLEAU` sheet in excel
- support xml template definition in metasheet which can ignore the contents in xml
- support `Sep` definition in metasheet which allows users to define whatever delimiter they like

Bug fixes:
- support list type instead of only keyed list
- support original name defined in xml file instead of auto-converted camel-case name


wenchy (approved):
LGTM

PR #43, opened 2023-03-21
feat: if param column of a union field not exist, use the default value rather than raising errors

Kybxd:
A union message may require too many field, but only Param1 is defined in EXCEL files. In this case, use default value for params not in EXCEL is better than raising an error.

Example: The "Fuben" TaskTarget requries 8 fields, but the EXCEL only has Param1, now Param2 to Param8 are filled with the default value 0.
企业微信截图_16793864916543
企业微信截图_16793866091741

### DIFF
--- a/cmd/tableauc/main.go
+++ b/cmd/tableauc/main.go
@@ -15,7 +15,7 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-const version = "0.5.7"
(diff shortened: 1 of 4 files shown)


wenchy:
This should not be allowed,  as it is not consistent with the design principle: each field should be explicit lypresent, otherwise rasie an missing field error.

PR #46, opened 2023-03-29
feat: support custom field json name

wenchy:
- Add new field prop option: `json_name`, to support custom protobuf field json name, refer [google/protobuf/descriptor.proto FieldDescriptorProto#json_name](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/descriptor.proto#L235).
- Add two tableauc config options:
  - option `UseProtoNames` uses proto field name instead of lowerCamelCase name in JSON field names.
  - option `UseEnumNumbers` emits enum values as numbers.


huieric (approved):
LGTM

PR #65, opened 2023-05-15
confgen: export deterministic JSON/Text/Bin

### LINKED ISSUE #61: confgen: export deterministic JSON/Text/Bin
## Problems

Protobuf `map` type is a hashmap, so elements are not ordered and non-deterministic.

- [encoding/protojson#Marshal](https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson#Marshal)
Marshal writes the given proto.Message in JSON format using default options. **Do not depend on the output being stable**. It may change over time across different versions of the program.
- [encoding/prototext#Marshal](https://pkg.go.dev/google.golang.org/protobuf/encoding/prototext#Marshal)
Marshal writes the given proto.Message in textproto format using default options. **Do not depend on the output being stable**. It may change over time across different versions of the program.
- [encoding/protowire#Marshal](https://pkg.go.dev/google.golang.org/protobuf/proto#Marshal)

### Related issues
- https://github.com/golang/protobuf/issues/1373
- https://github.com/bufbuild/connect-go/pull/460
- https://protobuf.dev/reference/go/faq/#unstable-json
- https://protobuf.dev/reference/go/faq/#unstable-text

## Solutions
- [x] JSON: to gain some degree of output stability, we recommend running the output through a JSON formatter.
- [x] Text: To obtain some degree of stability, we recommend passing the output of prototext through the [txtpbfmt](https://github.com/protocolbuffers/txtpbfmt) program. The formatter can be directly invoked in Go using [parser.Format](https://pkg.go.dev/github.com/protocolbuffers/txtpbfmt/parser?tab=doc#Format).
- [x] Bin: set marshal option `Deterministic` as `true`. See [MarshalOptions](https://github.com/protocolbuffers/protobuf-go/blob/v1.30.0/proto/encode.go#L21).

wenchy:
- close #61


internal/confgen/mexporter/marshal.go
```diff
@@ -0,0 +1,112 @@
...
+		return nil, err
+	}
+	if options.Pretty {
+		prettyJSON := new(bytes.Buffer)
+		if err := json.Indent(prettyJSON, stableJSON.Bytes(), "", "    "); err != nil {
+			return nil, err
+		}
+		return prettyJSON.Bytes(), nil
+	}
+	return stableJSON.Bytes(), nil
```
huieric, line 77: Add tests to cover cases not using `Pretty` option.

internal/confgen/mexporter/mexporter.go
```diff
@@ -59,19 +57,25 @@ func (x *messageExporter) export(fmt format.Format) error {
 	switch fmt {
 	case format.JSON:
 		filename += format.JSONExt
-		out, err = x.marshalToJSON()
+		options := &MarshalOptions{
+			Pretty:          x.outputOpt.Pretty,
+			EmitUnpopulated: x.outputOpt.EmitUnpopulated,
+			UseProtoNames:   x.outputOpt.UseProtoNames,
+			UseEnumNumbers:  x.outputOpt.UseEnumNumbers,
+		}
```
huieric, line 65: Add tests to cover these new lines.

internal/confgen/mexporter/marshal.go
wenchy: Done

internal/confgen/mexporter/mexporter.go
wenchy: Done

PR #83, opened 2023-11-13
feat(load): add option Paths

wenchy:
`Paths` maps each messager name to a corresponding config file path.
If a messager name is existed, then the messager will be parsed from the config file directly.
NOTE: only JSON, bin, and text formats are supported.


Kybxd (approved):
LGTM

PR #84, opened 2023-12-28
feat(xml): add support for merger

Kybxd:
In previous if you try to merge xmls, you need to add metasheet to each cloned workbook to make sure xml data is converted to 2D rows correctly. Now you only need to add metasheet to the primary workbook, and when confgen works on cloned workbooks, the metasheet of primary workbook is used


huieric (approved):
LGTM

PR #110, opened 2024-07-13
fix: IEEE-754 64-bit floats (float64 in Go) can only accurately represent integers in the range [-2^53,2^53]

wenchy:


Kybxd (approved):
LGTM

PR #117, opened 2024-08-03
XML: improve XML importer and document parser

wenchy:
- fix: keep node's children order stable
- chore: remove deprecated XML test cases in test/dev


soukengo (approved):
LGTM

PR #166, opened 2024-11-26
feat(patch): add loading mode option

Kybxd:


load/options.go
```diff
@@ -132,3 +139,13 @@ func PatchDirs(dirs ...string) Option {
...
+
+// IgnoreMainFile signifies whether to ignore the main file when
+// parsing a sheet with patch.
+//
+// NOTE: only JSON, Bin, and Text formats are supported.
+func IgnoreMainFile() Option {
+	return func(opts *Options) {
+		opts.IgnoreMainFile = true
+	}
+}
```
wenchy, line 151: We need to support three kinds of loading mode:
1. Default: Load all related files (both main and patch files)
2. OnlyMain: Only load the main file
3. OnlyPatch:  Only load the patch files

```go
type Options struct {
	// Mode specifies the loading mode for config patching.
	//
	// Default: ModeDefault.
	Mode Mode
}
type Mode int

const (
	ModeDefault   = iota // Load all related files
	ModeOnlyMain         // Only load the main file
	ModeOnlyPatch        // Only load the patch files
)

// SetMode specifies the loading mode for config patching.
//
// NOTE: only JSON, Bin, and Text formats are supported.
func SetMode(mode Mode) Option {
	return func(opts *Options) {
		opts.Mode = mode
	}
}
```

wenchy (changes requested):

PR #180, opened 2025-02-26
feat: add go.mod for tableauc

Kybxd:


cmd/tableauc/go.mod
```diff
@@ -0,0 +1,42 @@
+module github.com/tableauio/tableau/cmd/tableauc
+
+go 1.21
```
wenchy, line 3: Currently, use `go 1.20` instead.

wenchy (changes requested):

wenchy (approved):

PR #183, opened 2025-03-06
load(BREAKING)!: move Filter option to loader

Kybxd:
## ❗BREAKING CHANGE

In file *load/options.go*, we deleted the load option `Filter` for design consistency. Now you can find this option in the generated code in loader, and use this option in `tableau.NewHub`.

## Compatibility resolution
- Assign the filter func when initializing the hub

Old：
```go
hub := tableau.NewHub()
err := hub.Load(configDir, format.JSON,
    load.IgnoreUnknownFields(),
    load.Filter(yourCustomFilter),
)
```

New:
```go
hub := tableau.NewHub(tableau.Filter(yourCustomFilter))
err := hub.Load(configDir, format.JSON,
    load.IgnoreUnknownFields(),
)
```

### DIFF
--- a/load/load_test.go
+++ b/load/load_test.go
@@ -28,18 +28,6 @@ func TestLoad(t *testing.T) {
 		errcode string
 	}{
 		{
-			name: "load-origin",
-			args: args{
-				msg: &unittestpb.ItemConf{},
-				dir: "../testdata/",
-				fmt: format.CSV,
-				options: []Option{
-					Filter(nil), // just for test coverage
-				},
-			},
-			wantErr: false,
-		},
-		{
 			name: "load-origin-path-failed",
 			args: args{
 				msg: &unittestpb.ItemConf{},
--- a/load/load_test.go
+++ b/load/load_test.go
@@ -28,6 +28,16 @@ func TestLoad(t *testing.T) {
 		errcode string
 	}{
 		{
+			name: "load-origin",
+			args: args{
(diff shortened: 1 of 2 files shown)


load/options.go
```diff
@@ -55,6 +58,9 @@ type Options struct {
 	//
 	// Default: ModeDefault.
 	Mode LoadMode
+	// Validator enables the immutability validation of the loaded config, and
+	// specifies its interval and error handler.
+	Validator *Validator
```
wenchy, line 63: - Should define option at loader repo https://github.com/tableauio/loader, but not here.
- Naming: `Validator` -> `ImmutableChecker`

wenchy (changes requested):

load/load_test.go
```diff
@@ -27,18 +27,6 @@ func TestLoad(t *testing.T) {
...
 	}{
-		{
-			name: "load-origin",
-			args: args{
-				msg: &unittestpb.ItemConf{},
-				dir: "../testdata/",
-				fmt: format.CSV,
-				options: []Option{
-					Filter(nil), // just for test coverage
-				},
```
wenchy, line 38: Just delete options, but not the whole test case, which is for testing load-origin case.

wenchy (changes requested):

load/load_test.go
Kybxd: done

wenchy (approved):

PR #186, opened 2025-03-13
feat: extract common parser logic

Kybxd:


internal/protogen/parser.go
```diff
@@ -28,12 +23,12 @@ const (
 	listVarSuffix = "_list" // list variable name suffix
 )
 
-type bookParser struct {
+type sheetParser struct {
```
wenchy, line 26: Should not change this name, as it is really a book level parser, not sheet level.

wenchy (approved):

PR #191, opened 2025-03-18
fix(BREAKING!): move Sep and Subsep from ConfInputOption to HeaderOption

Kybxd:
`Sep` and `Subsep` are specified in proto files so that loadOrigin can work correctly.


internal/confgen/table_parser_test.go
```diff
@@ -546,168 +544,3 @@ func TestTableParser_parseDocumentMetasheet(t *testing.T) {
 		})
 	}
 }
-
-func TestParser_parseSimpleIncellMapWithGlobalSep(t *testing.T) {
```
wenchy, line 550: `field-level` and `sheet-level`, maybe support `book-level`

wenchy (changes requested):

internal/confgen/table_parser_test.go
Kybxd: done

wenchy (approved):

PR #201, opened 2025-03-25
union: support integration for oneof fields

Kybxd:


wenchy:
- complicated: subtle and tricky
- break the rule: what-you-see-is-what-you-get

PR #223, opened 2025-04-15
protogen(union): use CamelCase for union's oneof message

Kybxd:


internal/protogen/exporter.go
```diff
@@ -224,7 +224,7 @@ func (x *sheetExporter) exportUnion() error {
 		if len(field.Fields) == 0 {
 			x.g.P("    // No field bound to enum value: ", ename, ".")
 		} else {
-			x.g.P("    ", strings.TrimSpace(field.Name), " ", x.be.gen.strcaseCtx.ToSnake(field.Name), " = ", field.Number, `; // Bound to enum value: `, ename, ".")
+			x.g.P("    ", x.be.gen.strcaseCtx.ToCamel(field.Name), " ", x.be.gen.strcaseCtx.ToSnake(field.Name), " = ", field.Number, `; // Bound to enum value: `, ename, ".")
```
wenchy, line 227: This is a **breaking change**, need to take a deep think.

By default, we should use original `field.Name` (obey the rule: what-you-see-is-what-you-get).

wenchy (changes requested):

internal/protogen/exporter.go
wenchy: Resolved by another compatible and natural solution, see PR #224 .

wenchy:
Resolved by another compatible and natural solution, see PR #224 .
This PR should be closed.

PR #225, opened 2025-04-21
protogen: support processing all files in first-pass when only generate some specified files

### LINKED ISSUE #185: tableauc: auto analyze and include the predefined types defined in other files
For example, there are two excel files: 
- **Target.xlsx**: define the union type `Target` 
- **Task.xlsx**:  use the type `Target` defined in Target.xlsx

When execute `tableauc -c config.yaml Task.xlsx`, it should work nicely without error message: 
```
predefined type not found: .Target
```

wenchy:
- close #185

### DIFF
--- a/internal/protogen/protogen.go
+++ b/internal/protogen/protogen.go
@@ -108,17 +108,8 @@ func (gen *Generator) GenAll() error {
 		return err
 	}
 	// first pass
-	if len(gen.InputOpt.Subdirs) != 0 {
-		for _, subdir := range gen.InputOpt.Subdirs {
-			dir := filepath.Join(gen.InputDir, subdir)
-			if err := gen.generate(dir); err != nil {
-				return err
-			}
-		}
-	} else {
-		if err := gen.generate(gen.InputDir); err != nil {
-			return err
-		}
+	if err := gen.processFirstPass(); err != nil {
+		return err
 	}
 	// second pass
 	return gen.processSecondPass()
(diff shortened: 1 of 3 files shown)


internal/protogen/protogen.go
```diff
@@ -129,18 +120,50 @@ func (gen *Generator) GenWorkbook(relWorkbookPaths ...string) error {
...
+	log.Infof("%15s: processing all books", "first-pass")
+	if len(gen.InputOpt.Subdirs) != 0 {
+		for _, subdir := range gen.InputOpt.Subdirs {
+			dir := filepath.Join(gen.InputDir, subdir)
+			if err := gen.processDirFirstPass(dir); err != nil {
+				return err
+			}
+		}
+	} else {
+		if err := gen.processDirFirstPass(gen.InputDir); err != nil {
```
Kybxd, line 162: just use `else if`

internal/protogen/protogen.go
```diff
@@ -156,13 +179,13 @@ func (gen *Generator) processSecondPass() error {
 	for _, absPath := range absPaths {
 		absPath := absPath
 		eg2.Go(func() error {
```
Kybxd, line 181: `eg2` -> `eg`

Kybxd (changes requested):

internal/protogen/protogen.go
wenchy: fixed

internal/protogen/protogen.go
wenchy: more clean code improved

PR #230, opened 2025-04-23
protogen: add field number prop

### LINKED ISSUE #218: protogen: guarantee message field's Tag Number compatibility
See https://protobuf.dev/best-practices/dos-donts/
> Don’t Re-use a Tag Number
Never re-use a tag number. It messes up deserialization. Even if you think no one is using the field, don’t re-use a tag number. If the change was live ever, there could be serialized versions of your proto in a log somewhere. Or there could be old code in another server that will break.

To guarantee both backward and forward compatibility, the field's tag number should be fixed (specified explicitly), to avoid frequent changes (due to column additions/deletions) if auto generated by tableau protogen.

Kybxd:
- close #218


internal/protogen/exporter.go
```diff
@@ -293,7 +288,7 @@ func (x *sheetExporter) exportMessager() error {
 	return nil
 }
 
-func (x *sheetExporter) exportField(depth int, tagid int, field *internalpb.Field, prefix string) error {
+func (x *sheetExporter) exportField(depth int, tagid *int, field *internalpb.Field, prefix string) error {
```
wenchy, line 291: For Go, In/Out parameter resue is confusing, error-prone,  and not recommended.
Please return the tagid instead if you want to return next one (just like `cursor`).
```go
exportField(depth int, tagid int, field *internalpb.Field, prefix string) (int, error)
```

wenchy (changes requested):

proto/tableau/protobuf/tableau.proto
```diff
@@ -334,6 +334,12 @@ message FieldProp {
   //
   // TODO: use cases for more composite types.
   int32 cross = 15;
+  // Specify number of this field.
+  // All fields with number unspecified will use a sequence starting from 1 by default, 
+  // and users should guarantee this number do not conflict with the default sequence.
+  //
+  // For tables, assigning number to a cross cell map/list/struct field also works on its 1st sub-field.
```
Kybxd, line 341: Is this acceptable, or we may find another way to distinguish the `number` prop works on the list/map, or the first subfield of the list/map?

test/functest/proto/default/excel__fieldprop__field_prop.proto
```diff
@@ -21,6 +21,28 @@ message EquipConf {
...
+message FieldNumberProp {
+  option (tableau.worksheet) = {name:"FieldNumberProp"};
+
+  map<uint32, Item> item_map = 1 [(tableau.field) = {key:"ID" layout:LAYOUT_VERTICAL}];
+  message Item {
+    uint32 id = 1 [(tableau.field) = {name:"ID"}];
+    string name = 2 [(tableau.field) = {name:"Name"}];
+    int32 price = 10 [(tableau.field) = {name:"Price" prop:{number:10}}];
+    int32 number = 3 [(tableau.field) = {name:"Number"}];
+    Decompose decompose = 20 [(tableau.field) = {name:"Decompose" prop:{number:20}}];
```
wenchy, line 33: 1. Should remove the prop `number:10` as it is already used to specify tag id, no useless to keep it.
2. If specified prop `number`, then the other auto generated tag id MUST not use it again in the same nesting level. Add test case to cover it.

proto/tableau/protobuf/tableau.proto
wenchy: - I think we'd better distinguish them. Add new prop `subnumber` (just like`subsep`) to specify first sub-field's number of the list/map element.
- Also need to improve comment below.
```protobuf
  // Specify field number (unique numbered tag).
  // All fields without numbers specified will use a sequence starting from 1 by default.
  // Besides, auto generated field numbers will not reuse explicitly specified field numbers.
  //
  // Refer to [protobuf: Assigning Field Numbers](https://protobuf.dev/programming-guides/proto3/#assigning)
  int32 number = 16;
  // Specify first sub-field's number of the list-element/map-value/struct.
  int32 subnumber = 17;
```

wenchy (changes requested):

internal/protogen/exporter.go
Kybxd: done

proto/tableau/protobuf/tableau.proto
Kybxd: We fixed the field number of message's first field to 1. So it's no need to consider `subnumber` now.

test/functest/proto/default/excel__fieldprop__field_prop.proto
Kybxd: 1. done
2. a new algorithm is introduced to guarantee auto generated field numbers do not conflict with specified ones.

PR #231, opened 2025-04-24
feat: support sequence in auto-merge map keys

### LINKED ISSUE #163: FieldProp(sequence ): support sequence check even when parent same-key auto merge
| ID | SectionID | Stage | Desc |
|--------|-----|------|----| 
| map<uint32, Chapter>| map<uint32, Section>\|{sequence:1}| [Stage]uint32| string |
| Chapter ID | Section ID |  Stage | Desc |
| 1001| 1 | 1|  desc1-1 |
| 1001| 1 | 2|  desc1-2 |
| 1001| 2 | 1 | desc2-1 |
| 1001| 2 | 2 | desc2-2 |
| 1001| 2 | 3 | desc2-3 |
| 1001| 3 | 1 | desc3-1 | 

Based on PR #205, this feature should be easy to implement.

Kybxd:
- close #163


internal/confgen/parser.go
```diff
@@ -553,8 +557,9 @@ func (sp *sheetParser) checkMapKeyUnique(field *Field, reflectMap protoreflect.M
 	return nil
 }
 
-func (sp *sheetParser) checkListKeyUnique(field *Field, md protoreflect.MessageDescriptor, keyData string) error {
+func (sp *sheetParser) checkListKeyUnique(field *Field, reflectList protoreflect.List, keyData string) error {
 	// TODO: we need to cache key field descriptor for reuse, to avoid each loop to find it when parse table rows or document nodes.
+	md := reflectList.NewElement().Message().Descriptor()
```
wenchy, line 562: For higher performance, we'd better NOT create new element for getting **MessageDescriptor**, it is costy. Just use the provided parameter `md`.

internal/confgen/parser.go
```diff
@@ -608,18 +613,73 @@ func (sp *sheetParser) deduceKeyUnique(fieldLayout tableaupb.Layout, md protoref
...
+	}
+	keyField := sp.parseFieldDescriptor(fd)
+	defer keyField.release()
+	if !fieldprop.RequireSequence(keyField.opts.Prop) {
+		// do not require sequential
+		return nil
+	}
+	seq := keyField.opts.Prop.GetSequence()
+	len := int64(reflectMap.Len())
+	val, err := strconv.ParseInt(keyData, 10, 64)
```
wenchy, line 631: 1. Too many same code lines in the two func: `checkMapKeySequence` and `checkListKeySequence`, please encaplulate the same code lines to one func.
2. `strconv.ParseInt(keyData, 10, 64)` only support `int64`, but will not support `uint64` when its value is out-bound-of `int64`.

wenchy (changes requested):

internal/confgen/parser.go
Kybxd: done

internal/confgen/parser.go
Kybxd: 1. done
2. no need to support since values > max int64 is hardly used. `tableau.proto` also declares sequence as int64

wenchy: Yes, tableau.proto declares sequence as `int64`
https://github.com/tableauio/tableau/blob/4b09f7c4de471723b7569812e8a53ab5dac42d33/proto/tableau/protobuf/tableau.proto#L278-L280

wenchy (approved):

PR #252, opened 2025-05-28
protogen(naming): add support to check the messager naming pattern

### LINKED ISSUE #250: protogen(naming): add support to check the messager naming pattern
Add option `naming.messagerPattern` to protogen part in `config.yaml` to check messager naming pattern. For example, each messager must ends with `Conf` suffix.

Kybxd:
- close #250

### DIFF
--- a/internal/protogen/exporter.go
+++ b/internal/protogen/exporter.go
@@ -2,6 +2,7 @@ package protogen
 
 import (
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/emirpasic/gods/sets/treeset"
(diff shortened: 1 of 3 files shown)


internal/protogen/exporter.go
```diff
@@ -280,6 +284,9 @@ func (x *sheetExporter) exportUnion() error {
 
 func (x *sheetExporter) exportMessager() error {
 	// log.Debugf("workbook: %s", x.ws.String())
+	if x.be.messagerPatternRegexp != nil && !x.be.messagerPatternRegexp.MatchString(x.ws.Name) {
```
wenchy, line 287: Cannot check `messagerPatternRegexp != nil`, as MustCompile will not return nil, see https://pkg.go.dev/regexp#MustCompile

Instead, we should check `x.be.gen.InputOpt.MessagerPattern != ""`

Or maybe empty regex string wil match all patterns. So no need to check empty or nil.

wenchy (changes requested):

internal/protogen/exporter.go
Kybxd: This is to prevent panics in unittest

wenchy (approved):

PR #259, opened 2025-06-19
confgen(table): add a specifier to ingore some rows

### LINKED ISSUE #256: confgen(IgnoreRows): add a specifier to ignore some rows of data
For example, we can add a new specifier `#IGNORE` to prevent `ID=2` row from exporting, so it' will not be generated to JSON/Text/Bin file.
The `#IGNORE` column cell should be parsed by [strconv.ParseBool](https://pkg.go.dev/strconv#ParseBool), which can accept 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False. 

| ID                | Name        | Desc                          | #IGNORE |
| ----------------- | ----------- | ----------------------------- | - |
| map<uint32, Item> | string      | string                        ||
| Item’s ID         | Item’s name | Item’s desc                   ||
| 1                 | Apple       | A kind of delicious fruit.    ||
| 2                 | Orange      | A kind of sour fruit.         |true|
| 3                 | Banana      | A kind of calorie-rich fruit. ||

```json
{
    "itemMap": {
        "1": {
            "id": 1,
            "name": "Apple",
            "desc": "A kind of delicious fruit."
        },
        "3": {
            "id": 3,
            "name": "Banana",
            "desc": "A kind of calorie-rich fruit."
        }
    }
}
```

Kybxd:
- close #256


internal/importer/book/cell.go
```diff
@@ -262,3 +263,13 @@ func (r *RowCells) GetCellCountWithPrefix(prefix string) int {
 	}
 	return size
 }
+
+func (r *RowCells) Ignored() bool {
+	if cell, err := r.Cell("#IGNORE", false); err == nil {
+		if ok, _ := xproto.ParseBool(cell.Data); ok {
```
wenchy, line 269: DO NOT ignore func returned error!
If encounterd error, the process should crash and report the error.

internal/x/xproto/value.go
```diff
@@ -288,6 +280,17 @@ func GetFieldDefaultValue(fd pref.FieldDescriptor) string {
 	return ""
 }
 
+func ParseBool(value string) (bool, error) {
```
wenchy, line 283: For public func, should add func comment.

wenchy (changes requested):

internal/importer/book/cell.go
Kybxd: This is meant to accept any values on `#IGNORE` column, but only those where `xproto.ParseBool` returns true are recognized as ignoring this line.

internal/x/xproto/value.go
Kybxd: done

internal/x/xproto/value.go
```diff
@@ -288,6 +280,15 @@ func GetFieldDefaultValue(fd pref.FieldDescriptor) string {
 	return ""
 }
 
+// ParseBool parses bool value from cell data. Floats with a decimal part of 0 are treated as integers.
+func ParseBool(value string) (bool, error) {
+	// trim integer boring suffix matched by regexp `.0+$`
+	if matches := types.MatchBoringInteger(value); len(matches) > 1 {
+		return strconv.ParseBool(matches[1])
+	}
```
wenchy, line 288: Any reason to do this special converting?

wenchy (changes requested):

internal/x/xproto/value.go
Kybxd: This is the original logic in `purifyInteger`, which is deleted by this merge request.

internal/importer/book/cell.go
wenchy: It's subtle if user fills a notional string, e.g.: "x", "ignore". Then if tableauc does not report error, but treats it as not **ignored**, it will be confusing and introduce potential bugs.

internal/x/xproto/value.go
wenchy: No need to do this again. It is a history problem.

Kybxd: done

PR #273, opened 2025-07-22
feat: remove useless withCode

Kybxd:
When a messager name fails to match the required pattern, error messages of protogen contains a suspicious `: -1` after SheetName. This is because `xerrors.Errorf` creates an error with error code `-1` whose error message is like `-1: |Reason: messager <name> does not match pattern <pattern>`, and when it's wrapped with kvs, it becomes `SheetName: <sheetname>: -1: |Reason: messager <name> does not match pattern <pattern>`. `xerrors.Desc` splits the error message and parses its sheetname as `<sheetname>: -1`.

企业微信截图_17531686753245

### DIFF
--- a/xerrors/errors.go
+++ b/xerrors/errors.go
@@ -1,8 +1,7 @@
 //   Error handling model:
 // 			1. cause error(nil means no cause) is wrapped by base error with caller stack
 //          2. all errors contain only one caller stack
-//          3. withMessage is an error whitch has a message, which could be infinitely
-// 		       nested with each other
+//          3. withMessage is an error with a message, which could be infinitely nested with each other
 //
 //                                 +---------+
 //                                 |  cause  |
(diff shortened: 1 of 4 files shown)


xerrors/errors.go
```diff
@@ -1,9 +1,8 @@
 //   Error handling model:
 // 			1. cause error(nil means no cause) is wrapped by base error with caller stack
 //          2. all errors contain only one caller stack
-//          3. withCode is an error which has a code
-//          4. withMessage is an error whitch has a message
-//          5. withCode and withMessage could be infinitely nested and nested with each other
+//          3. withMessage is an error whitch has a message, which could be infinitely
+// 		       nested with each other
```
wenchy, line 5: - typo: `whitch` -> `with`
- fix line format

wenchy (changes requested):

xerrors/errors.go
Kybxd: done

wenchy (approved):

PR #287, opened 2025-08-19
feat(BREAKING!): update load options

Kybxd:


load/options.go
```diff
@@ -226,8 +245,11 @@ func SubdirRewrites(subdirRewrites map[string]string) Option {
 // NOTE: only JSON, Bin, and Text formats are supported.
 //
 // Deprecated: use [WithMessagerOptions] instead.
-func Paths(paths map[string]string) Option {
-	return func(opts *Options) {
+func Paths(paths map[string]string) LoadOption {
```
wenchy, line 248: Now, we can remove this deprecated func.

load/options.go
```diff
@@ -243,8 +265,11 @@ func Paths(paths map[string]string) Option {
 // NOTE: only JSON, Bin, and Text formats are supported.
 //
 // Deprecated: use [WithMessagerOptions] instead.
-func PatchPaths(paths map[string][]string) Option {
-	return func(opts *Options) {
+func PatchPaths(paths map[string][]string) LoadOption {
```
wenchy, line 268: Now, we can remove this deprecated func.

load/options.go
```diff
@@ -19,14 +19,14 @@ type BaseOptions struct {
 	// If the name is "Local", LoadLocation returns Local.
 	//
 	// Default: "Local".
-	LocationName string
+	LocationName *string
```
wenchy, line 22: Empty can indicate the field presence, no need to change it to pointer.

wenchy (changes requested):

load/options.go
```diff
@@ -154,67 +186,54 @@ func ParseMessagerOptionsFromOptions(o *Options, name string) *MessagerOptions {
...
-			LocationName: "Local",
-			Mode:         ModeAll,
-			ReadFunc:     os.ReadFile,
-			LoadFunc:     LoadMessager,
-		},
-		MessagerOptions: map[string]*MessagerOptions{},
-	}
-}
+// LoadOptions is the functional option type.
+type LoadOption func(*LoadOptions)
```
wenchy, line 193: Maybe there is no need to change naming, as it is aready in `load` package:
- Options-> LoadOptions
- Option -> LoadOption

wenchy (changes requested):

load/options.go
Kybxd: The `LocationName` option is parsed by `time.LoadLocation`, where `""` equals `"UTC"`. So only pointer explicitly indicates the field presence. If we do not use a pointer, we cannot distinguish whether an empty string is because the user doesn't specify this option (they wants `"Local"` by default), or because they specifies `load.LocationName("")` (they wants `""` which equals `"UTC"`).

```go
// go/src/time/zoneinfo.go
func LoadLocation(name string) (*Location, error) {
	if name == "" || name == "UTC" {
		return UTC, nil
	}
	if name == "Local" {
		return Local, nil
	}
	...
}
```

load/options.go
Kybxd: Keep consistent with cpp implementations is better for subsequent maintenance.

load/options.go
Kybxd: Should `func Load(msg proto.Message, dir string, fmt format.Format, options ...LoadOption) error` also be removed?

load/options.go
wenchy: C++ or other programming languages can use `load` namespace:

For example:
```cpp
namespace load {

struct Options {
  ...
}

struct MessagerOptions {
  ...
}

} // namespace load
```

load/load.go
```diff
@@ -37,46 +38,48 @@ func LoadMessagerInDir(msg proto.Message, dir string, fmt format.Format, opts *M
...
-	content, err := opts.ReadFunc(path)
+	content, err := opts.GetReadFunc()(path)
 	if err != nil {
 		return xerrors.Wrapf(err, "failed to read file: %v", path)
 	}
 	return Unmarshal(content, msg, path, fmt, opts)
 }
 
-func loadWithPatch(msg proto.Message, path string, fmt format.Format, patch tableaupb.Patch, opts *MessagerOptions) error {
+func LoadMessagerWithPatch(msg proto.Message, path string, fmt format.Format, patch tableaupb.Patch, opts *MessagerOptions) error {
```
wenchy, line 68: There is no need to export this func?

load/load.go
```diff
@@ -136,6 +139,9 @@ func loadWithPatch(msg proto.Message, path string, fmt format.Format, patch tabl
 //
 // NOTE: only output formats (JSON, Bin, Text) are supported.
 func Unmarshal(content []byte, msg proto.Message, path string, fmt format.Format, opts *MessagerOptions) error {
+	if opts == nil {
+		opts = &MessagerOptions{}
+	}
```
wenchy, line 144: Can be removed?

wenchy (changes requested):

PR #302, opened 2025-09-10
fix(timestamppb): converting output to local timezone safely

Kybxd:


store/jsonutil.go
```diff
@@ -18,7 +21,14 @@ func formatTimestamp(ts string, loc *time.Location) string {
...
 	}
 	localTime := t.In(loc)
-	return localTime.Format(time.RFC3339Nano)
+	formatted := localTime.Format(time.RFC3339Nano)
+	if err := protojson.Unmarshal(
+		fmt.Appendf(nil, `"%s"`, formatted),
+		&timestamppb.Timestamp{},
+	); err != nil {
+		return ts // return the original string if parsing fails
+	}
```
wenchy: It's time consuming, just compare it with `MaxTime()` with **-14 hours** (as the offsets range from UTC−12:00 to UTC+14:00, refer to https://en.wikipedia.org/wiki/Time_zone), to decide whether or not to format it with timezone? 
```go
// MaxTime returns the local time upper limit of [timestamppb].
//
// NOTE: In local time zone [time.Local].
//
// [timestamppb]: https://github.com/protocolbuffers/protobuf/blob/v31.1/src/google/protobuf/timestamp.proto#L133
func MaxTime() time.Time {
	return time.Date(9999, 12, 31, 23, 59, 59, 0, time.UTC).Local()
}
```

wenchy (changes requested):

wenchy (approved):

PR #312, opened 2025-11-03
refactor(ecode): add codegen and improve testability

wenchy:
- add codegen for ecode, and improve ecode testability
- fix xerrors wrap format bugs, and add more unit test cases to cover
- improve i18n package

### DIFF
--- a/internal/tools/cmd/ecode/main.go
+++ b/internal/tools/cmd/ecode/main.go
@@ -111,7 +111,7 @@ func loadConfig(path string) Config {
 func loadOtherLangConfigs(path string) map[string]Config {
 	otherLangConfigs := map[string]Config{} // lang path -> Config
 	dir := filepath.Dir(path)
-	filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
+	err := filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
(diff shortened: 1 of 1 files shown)


xerrors/errors.go
```diff
@@ -97,48 +101,60 @@ func (w *withMessage) Unwrap() error { return w.cause }
...
+		if w.cause != nil {
+			cause := fmt.Sprintf("%s", w.cause)
+			if cause != "" {
+				content += sep + cause
+			}
+		}
+		format := "%s"
+		switch verb {
+		case 's':
+			format = "%s"
```
Kybxd: Duplicate with default format `"%s"`

internal/localizer/i18n/config/ecode/en.yaml
```diff
@@ -104,23 +157,38 @@ E2022:
   desc: sub-field's value not unique in map values or list elements
   text: value "{{.Key}}" of sub-field "{{.FieldName}}" duplicates in map values or list elements
   help: ensure value of sub-field "{{.FieldName}}" is unique in map values or list elements
+  fields:
+    - FieldName: string
```
Kybxd: Just define `FieldName` type as `protoreflect.Name` to avoid type conversions in code?

PR #315, opened 2025-11-14
feat: recognize metasheet noteline and improve datetime error message

### LINKED ISSUE #278: proto: support to set noteline in metasheet
noteline can only be set by config.yaml currently, metasheet also need this.

### LINKED ISSUE #280: ecode: improve error message of invalid datetime format for time.RFC3399
Since #257 has supported `time.RFC3339` format, the error message should also be improved.
en
https://github.com/tableauio/tableau/blob/719d39844ff2afb37b1e60371cbef2240875adcb/internal/localizer/i18n/config/ecode/en.yaml#L46
zh
https://github.com/tableauio/tableau/blob/719d39844ff2afb37b1e60371cbef2240875adcb/internal/localizer/i18n/config/ecode/zh.yaml#L46

wenchy:
- close #278
- close #280

### DIFF
--- a/internal/localizer/i18n/config/ecode/en.yaml
+++ b/internal/localizer/i18n/config/ecode/en.yaml
@@ -72,7 +72,7 @@ E2006:
 E2007:
   desc: invalid datetime format
   text: '"{{.Value}}" is invalid datetime format, {{.Error}}'
-  help: 'follow datetime format: "yyyy-MM-dd HH:mm:ss", e.g.: "2020-01-01 01:00:00"'
+  help: 'follow datetime format: "yyyy-MM-dd HH:mm:ss" or RFC3339, e.g.: "2020-01-01 01:00:00" or "2020-01-01T01:00:00+08:00"'
 E2008:
   desc: invalid duration format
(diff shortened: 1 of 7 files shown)


internal/localizer/i18n/config/ecode/en.yaml
```diff
@@ -72,7 +72,7 @@ E2006:
 E2007:
   desc: invalid datetime format
   text: '"{{.Value}}" is invalid datetime format, {{.Error}}'
-  help: 'follow datetime format: "yyyy-MM-dd HH:mm:ss", e.g.: "2020-01-01 01:00:00"'
+  help: 'follow datetime format: "yyyy-MM-dd HH:mm:ss" or RFC3339, e.g.: "2020-01-01 01:00:00" or "2020-01-01T01:00:00+08:00"'
```
Kybxd: Replace `"yyyy-MM-dd HH:mm:ss"` with `DateTime`, which is consistent with `RFC3339`?

wenchy: DateTime is not a standard. So the pattern is more readable.

PR #332, opened 2025-12-26
feat: support generating multiple messagers for the same sheet

### LINKED ISSUE #328: Proposal: support generating multiple messagers for the same sheet
In some scenarios, we want to generate different columns for the same sheet, for example:
  - **ItemConf**: generate all columns of sheet Item
  - **ItemLiteConf**: only generate partial columns of sheet Item

So we need support generating multiple messagers for the same sheet. Then we need change `map<string, Metasheet> metasheet_map` to `repeated Metasheet sheet_list` 
https://github.com/tableauio/tableau/blob/8a16745f232f2be6628271334e587914609b779d/proto/tableau/protobuf/internal/metabook.proto#L12-L16

wenchy:
- close #328

Based on the **metasheet separation**, we can define two tableauc config files with different Metasheet name.

For example:

- config.yaml: use the default metasheet name `@TABLEAU`
- config2.yaml: set another metasheet name, such as `@TABLEAU2`

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -28,6 +28,7 @@ func init() {
 		driver: zapdriver.New(zap.NewDevelopmentConfig(), zap.AddCallerSkip(4)),
 	}
 	gOpts = &Options{}
+	atomicLevel = zap.NewAtomicLevelAt(zap.DebugLevel)
 }
 
 func Init(opts *Options) error {


internal/confgen/util.go
```diff
@@ -170,14 +176,14 @@ func buildWorkbookIndex(protoPackage, inputDir string, subdirs []string, subdirR
...
-					books: make(map[string]protoreflect.FileDescriptor),
+			rewrittenBookName := xfs.RewriteSubdir(workbook.Name, subdirRewrites)
+			if bookIndexes[rewrittenBookName] == nil {
+				bookIndexes[rewrittenBookName] = &primaryBookInfo{
+					books: make(map[string][]protoreflect.FileDescriptor),
 				}
 			}
-			bookIndexes[rewrittenWorkbookName].books[workbook.Name] = fd
-			// merger or scatter (only one can be set at once)
+			bookIndexes[rewrittenBookName].books[workbook.Name] = append(bookIndexes[rewrittenBookName].books[workbook.Name], fd)
```
Kybxd: Can we use `struct{rewrittenBookName, workbook.Name}` as mapkey to simplefy code?

Kybxd (changes requested):

internal/confgen/util.go
Kybxd: This seems hard but the second-level map key `workbook.Name` seems to be unused so we can just simplify it to a slice instead of map of slice?

wenchy: Resolved.

PR #343, opened 2026-01-08
refactor(BREAKING)!: consistent protoFiles parsing

Kybxd:
## ❗BREAKING CHANGE
- compiler: use `filepath.Rel` to make the path of `protoFiles` relative to `protoPaths`.
- protogen: `ProtoInputOption.ProtoFiles` accepts glob patterns relative to current working directory (consistent with confgen).
- protogen/confgen: all imports must be explicitly specified by `InputOption.ProtoFiles`, which guarantees predefined files not to import generated files.

## Compatibility resolution
- always specify the full path (relative to CWD) of all proto files imported in protogen and confgen, glob patterns are supported.

### DIFF
--- a/options/options.go
+++ b/options/options.go
@@ -216,9 +216,10 @@ type ConfOption struct {
 
 // Input options for generating conf files.
 type ConfInputOption struct {
-	// The proto paths are used to search for dependencies that are referenced in import
-	// statements in proto source files. If no import paths are provided then
-	// "." (current directory) is assumed to be the only import path.
+	// The proto paths are used to search for dependencies that are referenced
+	// in import statements in proto source files. If no import paths are
+	// provided then "." (current directory) is assumed to be the only import
+	// path.
 	//
 	// Default: nil.
 	ProtoPaths []string `yaml:"protoPaths"`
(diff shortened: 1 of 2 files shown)


internal/x/xproto/protoc/compiler.go
```diff
@@ -43,27 +51,42 @@ func NewFiles(protoPaths []string, protoFiles []string, excludedProtoFiles ...st
 		for _, match := range matches {
 			cleanSlashPath := xfs.CleanSlashPath(match)
 			if !parsedExcludedProtoFiles[cleanSlashPath] {
-				for _, protoPath := range protoPaths {
-					cleanProtoPath := xfs.CleanSlashPath(protoPath) + "/"
-					cleanSlashPath = strings.TrimPrefix(cleanSlashPath, cleanProtoPath)
-				}
-				parsedProtoFiles = append(parsedProtoFiles, cleanSlashPath)
+				rel := rel(cleanSlashPath, cleanSlashProtoPaths)
+				parsedProtoFiles[cleanSlashPath] = rel
```
wenchy: The excludedProtoFiles should be tested in unittest.

options/options.go
```diff
@@ -222,10 +225,8 @@ type ConfInputOption struct {
 
 	// The files to be parsed to generate configurations.
 	//
-	// NOTE:
-	//  - Recognize "*.proto" pattern if not set (value is nil).
```
wenchy: Well, the default behavior is ok, and should not be removed. It's useful for simple `tableauc` job without `config.yaml`. For example, the https://tableauio.github.io/docs/prologue/quick-start/. If removed, it will not work well.

wenchy (changes requested):

internal/x/xproto/protoc/compiler.go
Kybxd: done

wenchy (changes requested):

PR #351, opened 2026-02-04
fix(fieldprop): check presence for incell struct/list/map/union

### LINKED ISSUE #325: fieldprop(present): `present:true` not works for message field
```protobuf
message ItemConf {
  option (tableau.worksheet) = {name:"ItemConf "};

  uint32 id= 1 [(tableau.field) = {name:"ID"}]; 
  Effect effect = 2 [(tableau.field) = {name:"Effect" prop:{present:true}}]; // not works if this field not found in excel
  message Effect {
    int32 value = 1 [(tableau.field) = {name:"Value"}]; 
    int32 price = 2 [(tableau.field) = {name:"Price"}]; 
  }
}
```

Kybxd:
- close #325

### DIFF
--- a/internal/confgen/parser.go
+++ b/internal/confgen/parser.go
@@ -756,10 +756,11 @@ func (p *sheetParser) checkSubFieldProp(field *Field, cardPrefix string, newValu
 
 func (p *sheetParser) parseIncellList(field *Field, list protoreflect.List, cardPrefix string, elemData string) (present bool, err error) {
 	if elemData == "" {
-		// check presence
 		if err := fieldprop.CheckPresence(field.opts.Prop, false); err != nil {
 			return false, err
 		}
+		// For fixed size list, an empty cell means an empty list element,
+		// So here we should continue to parse, but not return.
 	}
 	splits := strings.Split(elemData, field.sep)
 	return p.parseListElems(field, list, cardPrefix, field.subsep, splits)


internal/confgen/parser.go
```diff
@@ -414,6 +414,10 @@ func (p *sheetParser) parseIncellMap(field *Field, reflectMap protoreflect.Map,
 //   - map<int32, EnumType>
 func (p *sheetParser) parseIncellMapWithSimpleKV(field *Field, reflectMap protoreflect.Map, cellData string) (err error) {
 	if cellData == "" {
+		// check presence
+		if err := fieldprop.CheckPresence(field.opts.Prop, false); err != nil {
+			return err
+		}
 		return nil
```
wenchy: Clean code, just `return fieldprop.CheckPresence(field.opts.Prop, false)`

internal/confgen/table_parser.go
```diff
@@ -631,7 +631,7 @@ func (p *tableParser) parseUnionField(field *Field, msg protoreflect.Message, r
 	if field.opts.Span == tableaupb.Span_SPAN_INNER_CELL {
 		// incell union
 		if cell, err = r.Cell(newPrefix, p.IsFieldOptional(field)); err == nil {
-			present, err = p.parseIncellUnion(structValue, cell.Data, field.opts.GetProp().GetForm())
+			present, err = p.parseIncellUnion(field, structValue, cell.Data)
```
wenchy: Please add test case to cover the error condition.

For example: https://github.com/tableauio/tableau/blob/v0.15.0/internal/confgen/table_parser_test.go#L344

wenchy (changes requested):

internal/confgen/parser.go
Kybxd: done

wenchy (approved):

PR #352, opened 2026-02-04
fix(protogen): use strings.LastIndex to trim first field name from predefined struct name

### LINKED ISSUE #314: confgen(union): generate failed if union name contains `Type`
> Refer to https://tableauio.github.io/docs/excel/union/#predefined-union-in-map

The error is : **error[E2014]: sheet column not found**

For example: 

| ID               | TargetTypeType                  | TargetTypeField1    | TargetTypeField2     | TargetTypeField3        | Progress |
| ---------------- | --------------------------- | --------------- | ---------------- | ------------------- | -------- |
| map<int32, Task> | {.Target}enum<.Target.Type> | union           | union            | union               | int32    |
| ID               | Target's type               | Target's field1 | Target's field2  | Target's field3     | Progress |
| 1                | PVP                         | 1               | 10               | Apple,Orange,Banana | 3        |
| 2                | PVE                         | 1,100,999       | 1,2,3            | 1:10,2:20,3:30      | 10       |
| 3                | Story                       | 1001,10         | 1:Apple,2:Orange | Fragrant:1,Sour:2   | 10       |
| 4                | Skill                       | 1               | 2                |                     | 8        |

Potential solution: should we just find the default `Type` field of union from end to begin  (current is begin to end)in column `TargetTypeType`?

TODO: more concrete example, as **confgen** generates the wrong union field.

Kybxd:
- close #314

### DIFF
--- a/internal/x/xproto/type_info.go
+++ b/internal/x/xproto/type_info.go
@@ -123,23 +123,31 @@ func extractTypeInfosFromMessage(md protoreflect.MessageDescriptor, typeInfos *T
 	}
 }
 
+// parseFirstFieldOptionName parses the first field option name of the message.
+//   - If the message is a union, return the name of the enum type field.
+//   - Else if the message has sub fields, return the name of the first field. Besides,
+//     if the first field's kind is message and its span is not inner cell,
+//     then recursively parse sub fields' option name and concat them.
+//   - Otherwise, return empty string.
 func parseFirstFieldOptionName(md protoreflect.MessageDescriptor) string {
-	firstFieldOptionName := ""
 	if IsUnion(md) {
 		desc := ExtractUnionDescriptor(md)
 		if desc != nil {
 			// union's first field is enum type field.
 			fieldOpts := proto.GetExtension(desc.Type.Options(), tableaupb.E_Field).(*tableaupb.FieldOptions)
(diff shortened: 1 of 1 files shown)


internal/protogen/table_parser.go
```diff
@@ -820,7 +820,7 @@ func (p *tableParser) parseStructField(field *internalpb.Field, header *tableHea
 					xerrors.KeyPBFieldOpts, desc.Prop.Text,
 					xerrors.KeyTrimmedNameCell, trimmedNameCell)
 			}
-			if index := strings.Index(trimmedNameCell, typeInfo.FirstFieldOptionName); index != -1 {
+			if index := strings.LastIndex(trimmedNameCell, typeInfo.FirstFieldOptionName); index != -1 {
```
wenchy: Well, if trim in reverse order, it will have a another problem if int32 (ItemID) in struct (Item) in struct (Target)
```
TargetItemItemID
{.Target}{.Item}int32
```

So the solution shoud be:
- If the struct's first field is not struct type, then trim in reverse order
- Otherwise, trim in forward order

Please also add test cases to cover the logic.

Kybxd: `FirstFieldOptionName` is parsed recursively now. New test cases added.

PR #354, opened 2026-02-05
fix(protogen): use fullpath when removing all protos except imports in outdir

Kybxd:


internal/protogen/util.go
wenchy: Please add test case for  prepareOutdir to cover the changed lines

wenchy (changes requested):

PR #355, opened 2026-02-05
feat(confgen): specify output formats for each messager

Kybxd:


internal/confgen/util.go
```diff
@@ -261,13 +261,22 @@ func loadProtoRegistryFiles(protoPackage string, protoPaths []string, protoFiles
 	return protoc.NewFiles(protoPaths, protoFiles, excludeProtoFiles...)
 }
 
+// parseOutputFormats parses the output formats of the specified message.
+func parseOutputFormats(msg proto.Message, opt *options.ConfOutputOption) []format.Format {
+	messagerName := string(msg.ProtoReflect().Descriptor().Name())
+	if formats, ok := opt.MessagerFormats[messagerName]; ok && len(formats) != 0 {
```
wenchy: Please add test case to cover MessagerFormats option

PR #395, opened 2026-04-15
feat(BREAKING)!: support aggregate for incell and horizontal map/list fields

Kybxd:
## Summary

Add a new `aggregate` field property (`FieldProp.aggregate`) that enables incell and horizontal map/list fields to collect elements across multiple rows (or columns) sharing the same parent key, instead of requiring identical values on every visit.

## Breaking Changes

Previously, incell map/list fields inside a vertical map (with duplicate parent keys) would silently collect values from different rows without any check. Now:

**Without `aggregate:true`**: different values across rows with the same parent key will report **E2023** (field value conflict across rows or columns). If your data intentionally has different incell values per row, you must add `|{aggregate:true}` to the type cell in the header.

#### Aggregate incell list
This also applies to standalone incell lists (not nested in a map). For example, a sheet with only one logical record spread across rows:

| Skills |
|---|
| `[]int32` |
| Hero skills |
| 1,2,3 |
| 4,5,6 |

Here `Skills` differs between rows but the message is the same (single record sheet). Previously silently collected. Now triggers **E2023**.

**Fix**:

| Skills |
|---|
| `[]int32\|{aggregate:true}` |
| Hero skills |
| 1,2,3 |
| 4,5,6 |

> **Note**: If all rows carry identical incell values (just repeated for readability), no change is needed — E2023 only triggers when values actually differ.

#### Aggregate incell list nested in map

If your Excel/CSV data has an incell list field that spreads values across multiple rows (the sheet's root message is parsed per row):

| ID | Tags |
|---|---|
| `map<uint32, Item>` | `[]int32` |
| Item ID | Item tags |
| 1 | 1,2,3 |
| 1 | 4,5,6 |
| 2 | 7,8 |

Previously `Tags` for Item 1 silently collected into `[1, 2, 3, 4, 5, 6]`. Now it triggers **E2023**.

**Fix**: add `|{aggregate:true}` to the type cell (row 2):

| ID | Tags |
|---|---|
| `map<uint32, Item>` | `[]int32\|{aggregate:true}` |
| Item ID | Item tags |
| 1 | 1,2,3 |
| 1 | 4,5,6 |
| 2 | 7,8 |

## Key Changes

- **New proto field**: `bool aggregate = 21` in `FieldProp` message
- **New error codes**:
  - `E2023`: field value conflict across rows or columns (when aggregate is not enabled and values differ)
  - `E2028`: duplicate elements in incell keyed-list
- **Simplified `E2005`**: narrowed from "map or keyed-list key not unique" to just "map key not unique"
- **Horizontal map/list aggregate**: `parseHorizontalMapField` and `parseHorizontalListField` now support `aggregate:true`, matching existing incell behavior
- **Removed dead code**: E2023 consistency checks in document parser (XML/YAML are tree-structured, fields are never visited twice)
- **Protogen fix**: `ExtractMapFieldProp` and `ExtractListFieldProp` now preserve `aggregate` for `LAYOUT_HORIZONTAL` (was only `LAYOUT_INCELL`)

## Behavior

| Field type | aggregate=false (default) | aggregate=true |
|---|---|---|
| incell map | E2023 if values differ across rows | Collect entries; E2005 on duplicate keys |
| incell list | E2023 if values differ across rows | Append elements unconditionally |
| incell keyed-list | E2023 if values differ across rows | Append elements; E2028 on duplicate keys |
| horizontal map | E2023 if values differ across rows | Collect entries; E2005 on duplicate keys |
| horizontal list | E2023 if values differ across rows | Append elements unconditionally |

## Test Coverage

- Unit tests: `TestTableParser_parseHorizontalAggregateMap`, `TestTableParser_parseHorizontalAggregateList`, `TestTableParser_parseIncellConsistencyWithoutAggregate`, `TestTableParser_parseIncellKeyedListWithDuplicateElements`, `TestTableParser_parseVerticalAggregationConsistency`
- Functional tests: `HAggregateMapInVMap`, `HAggregateListInVMap` (end-to-end protogen + confgen pipeline)


internal/confgen/table_parser.go
```diff
@@ -494,12 +536,40 @@ func (p *tableParser) parseIncellListField(field *Field, msg protoreflect.Messag
...
+						// check duplicate elems for scalar/enum keyed-list
+						for j := range presentList.Len() {
+							elemVal := presentList.Get(j)
+							if elemVal.Equal(newValue) {
+								return false, xerrors.WrapKV(xerrors.E2028(newValue), r.CellDebugKV(colName)...)
+							}
+						}
+					}
+					presentList.Append(newValue)
+				}
```
wenchy, line 583: 1. list aggregating, just append elems, no need to check uniqueness (if it is not keyed-list)
2. wellknown messages should also be considered

internal/localizer/i18n/config/ecode/zh.yaml
```diff
@@ -104,6 +104,10 @@ E2022:
   desc: sub-field's value not unique in map values or list elements
   text: map value 或 list element 子字段 "{{.FieldName}}" 的值 "{{.Key}}" 已存在
   help: 确保 map value 或 list element 子字段 "{{.FieldName}}" 的值唯一, 不能配置相同值
+E2023:
+  desc: inconsistent value in vertical aggregated row
+  text: 纵向聚合行的当前值 {{ quoteAny .Value }} 与首个出现的值 {{ quoteAny .FirstPresence }} 不相同
```
wenchy: 纵向聚合行 -> 垂直聚合

internal/localizer/i18n/config/ecode/zh.yaml
```diff
@@ -104,6 +104,10 @@ E2022:
   desc: sub-field's value not unique in map values or list elements
   text: map value 或 list element 子字段 "{{.FieldName}}" 的值 "{{.Key}}" 已存在
   help: 确保 map value 或 list element 子字段 "{{.FieldName}}" 的值唯一, 不能配置相同值
+E2023:
+  desc: inconsistent value in vertical aggregated row
```
wenchy: row -> rows

internal/localizer/i18n/i18n.go
```diff
@@ -14,7 +14,8 @@ import (
 
 // FuncMap contains the custom template functions used for rendering i18n templates.
 var FuncMap = template.FuncMap{
-	"quote": strconv.Quote,
+	"quote":    strconv.Quote,
+	"quoteAny": func(v any) string { return strconv.Quote(fmt.Sprint(v)) },
```
wenchy: Do not complicate it. Only one API `quote` is enough.

proto/tableau/protobuf/tableau.proto
```diff
@@ -367,6 +367,8 @@ message FieldProp {
   //
   // See https://github.com/bufbuild/protovalidate.
   string validate_message = 20;
+  // Whether this incell list/map field's value is aggregated vertically.
```
wenchy, line 370: Same as horizontally, should also be supported.

wenchy (changes requested):

internal/confgen/table_parser.go
Kybxd: 1. This is already implemented by current logic.
image

2. A better way to do this is to remove `field.fd.Kind() != protoreflect.MessageKind` check. In this situation, we can also check uniqueness for incell struct keyed-list.

internal/localizer/i18n/config/ecode/zh.yaml
Kybxd: done

internal/localizer/i18n/config/ecode/zh.yaml
Kybxd: done

internal/localizer/i18n/i18n.go
Kybxd: done

proto/tableau/protobuf/tableau.proto
Kybxd: I don't think it's a good idea to support aggregating a horizontal list vertically. This may look weird:

| WeaponID            | Attr1Name    | Attr1Value    | Attr2Name    | Attr2Value    | Level             | UpgradeCost                          |
| ------------------- | ------------ | ------------- | ------------ | ------------- | ----------------- | ------------------------------------ |
| map<uint32, Weapon> | [Attr]string | int32         | string       | int32         | map<int32, Level> | int32                                |
| Weapon's ID         | Attr1's name | Attr1's value | Attr2's name | Attr2's value | Weapon's Level    | Cost to upgrade weapon to next level |
| 1                   | Attack       | 100           | Defence      | 200           | 1                 | 10                                   |
| 1                   | Crit         | 30            | CritDamage   | 50            | 2                 | 20                                   |

wenchy: The design should be concistent and symmetry, Not only by visual.

test/functest/testdata/default/excel/fieldprop/FieldProp#FieldPropReferIgnored.csv
```diff
@@ -1,5 +1,5 @@
 ID,#IGNORE
-"[]int32|{refer:""EquipConf.ID""}",
+"[]int32|{refer:""EquipConf.ID"" aggregate:true}",
```
wenchy, line 2: This is a serious Breaking Change!

test/functest/testdata/default/excel/list/VerticalList#VerticalIncellUnionList.csv
```diff
@@ -1,5 +1,5 @@
 Target
-[]{union.Target}|{form:FORM_TEXT}
+[]{union.Target}|{form:FORM_TEXT aggregate:true}
```
wenchy, line 2: why need `aggregate:true` ?

test/functest/testdata/default/excel/nesting/KeyedListInKeyedList#VerticalKListInVerticalKList.csv
```diff
@@ -2,5 +2,5 @@ ID,Desc,PropID,PropNum
 [KeyedItem]<uint32>,string,[Prop]<uint32>,int32
 Item ID,Item desc,Prop ID,Prop num
 1,Apple,10,100
-1,Banana,11,110
+1,,11,110
 2,Orange,20,200
```
wenchy: The default behaviour should be: fill the same value as before (just like the same key filled).  Please also fix the other cases.

wenchy (changes requested):

proto/tableau/protobuf/tableau.proto
```diff
@@ -367,6 +367,8 @@ message FieldProp {
...
+  bool aggregate = 21;
```
wenchy, line 388: `merge` is a more accurate and intuitive word.

Image

test/functest/testdata/default/excel/list/VerticalList#VerticalIncellUnionList.csv
Kybxd: The list's layout is incell but is used vertically, so need `aggregate:true`.

test/functest/testdata/default/excel/nesting/KeyedListInKeyedList#VerticalKListInVerticalKList.csv
Kybxd: done

test/functest/testdata/default/excel/fieldprop/FieldProp#FieldPropReferIgnored.csv
Kybxd: updated to title & description.

wenchy (changes requested):

internal/confgen/parser.go
```diff
@@ -806,19 +806,15 @@ func (p *sheetParser) parseIncellList(field *Field, list protoreflect.List, card
...
-					return xerrors.WrapKV(xerrors.E2005(elemValue))
+					return xerrors.WrapKV(xerrors.E2028(elemValue))
 				}
 			}
 		}
 		// check list elem's sub-field prop
-		_, err = p.checkSubFieldProp(field, cardPrefix, elemValue)
-		if err != nil {
-			return err
-		}
```
wenchy, line 821: Why drop the code block?

Kybxd: This is 2 weeks ago code

wenchy (approved):

PR #396, opened 2026-04-16
refactor(wellknown): streamline Fraction comparison and drop LocalTime helper

Kybxd:
### Summary
Cleans up `proto/tableaupb/wellknown_util.go` by simplifying fraction comparison, promoting it to an idiomatic receiver-style method, and removing the redundant `LocalTime` helper.

### Changes

**1. `Fraction` comparison refactor**
- Introduce `(*Fraction).Cmp(cmp *Comparator) bool` as the canonical comparison entry point; the package-level `Compare` is kept as a thin wrapper for backward compatibility.

**2. Remove `LocalTime`**
- Delete `LocalTime(ts *timestamppb.Timestamp) time.Time` and its test.
- Callers can use `ts.AsTime().Local()` directly; the one-line wrapper added no real value and pulled `time` / `timestamppb` imports into this file unnecessarily.

### Compatibility
- `Compare(left, cmp)` signature is unchanged — existing callers keep working.
- `LocalTime` is removed; replace any usage with `ts.AsTime().Local()`.

### DIFF
--- a/proto/tableaupb/wellknown_util.go
+++ b/proto/tableaupb/wellknown_util.go
@@ -1,10 +1,6 @@
 package tableaupb
 
-import (
-	"time"
-
-	"google.golang.org/protobuf/types/known/timestamppb"
-)
+import "math/big"
 
 // NewFraction creates a new fraction.
 func NewFraction(num, den int32) *Fraction {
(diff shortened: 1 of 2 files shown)


wenchy:
### Code review

Found 2 issues:

1. `cmp.GetValue().AsRat()` panics when `Comparator.Value` is unset. In proto3, unset message fields return `nil` from getters. Calling `.AsRat()` on a nil `*Fraction` causes `GetDen()` to return 0, then `big.NewRat(0, 0)` panics with "division by zero". The original cross-multiplication code handled this gracefully without panicking.

https://github.com/tableauio/tableau/blob/da5b148cec6a1956b7a54f5a8c0342447edc1ead/proto/tableaupb/wellknown_util.go#L60-L64

2. `LocalTime()` is a public exported function being removed without a deprecation path. External consumers of the `tableaupb` package who depend on this function will break. Consider adding a `// Deprecated:` annotation first or documenting the removal in release notes.

https://github.com/tableauio/tableau/blob/da5b148cec6a1956b7a54f5a8c0342447edc1ead/proto/tableaupb/wellknown_util.go#L34-L36

Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

proto/tableaupb/wellknown_util.go
```diff
@@ -36,34 +32,33 @@ func NewIntegerComparator(sign Comparator_Sign, num int32) *Comparator {
...
-		panic("invalid compare operator")
-	}
+	return left.Cmp(cmp)
+}
+
+func (f *Fraction) AsRat() *big.Rat {
+	return big.NewRat(int64(f.GetNum()), int64(f.GetDen()))
+}
+
+var cmpResult = map[int]map[Comparator_Sign]bool{
```
wenchy:
```
  ┌─────────────────┬──────────────────────────────────────────────────────────────────────┬──────────────────────────────────────────────────────┐
  │     Aspect      │                      Previous (cross-multiply)                       │                 New (big.Rat + map)                  │
  ├─────────────────┼──────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────┤
  │ Allocations     │ 0 (stack-only int64 math)                                            │ 2+ heap allocations (big.Rat internals)              │
  ├─────────────────┼──────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────┤
  │ Performance     │ ~2 multiplications + 1 comparison                                    │ big.Rat normalization (GCD) + 2 map lookups          │
  ├─────────────────┼──────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────┤
  │ Overflow safety │ Theoretically can overflow int64 for large int32×int32               │ Arbitrary precision, no overflow                     │
  ├─────────────────┼──────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────┤
  │ Correctness     │ Safe for int32 inputs (max product = 2^31 × 2^31 = 2^62, fits int64) │ Always correct                                       │
  ├─────────────────┼──────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────┤
  │ Readability     │ Clear, self-contained switch                                         │ Compact but requires understanding the map structure │
  ├─────────────────┼──────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────┤
  │ Panic safety    │ Never panics (even on zero den)                                      │ Panics on zero denominator                           │
  └─────────────────┴──────────────────────────────────────────────────────────────────────┴──────────────────────────────────────────────────────┘
```

 The new version is worse on performance (heap allocations, GCD computation, map lookups vs. simple integer math) and worse on safety (panics on zero denominator).

  The overflow argument doesn't hold because int32 × int32 always fits in int64 (max value ~4.6×10^18 vs. max product ~4.6×10^18 — actually it's exactly at the edge: 2^31 × 2^31 = 2^62 which fits). So the original was already correct for int32
  inputs.

  The only real improvement is code compactness (fewer lines) and the addition of AsRat()/Cmp() as reusable public methods. But the map-based lookup is arguably less readable than the switch — a reader needs to mentally decode what
  cmpResult[-1][SIGN_LESS] means vs. seeing lval < rval directly.

  Summary: The refactor trades performance and safety for a more "clever" but less obvious implementation. The original switch was faster, safer, and equally readable.

wenchy (changes requested):

wenchy (approved):

PR #403, opened 2026-05-09
feat(protogen): preserve stable field numbers and auto-emit `reserved` statements on regeneration

### LINKED ISSUE #402: protogen: Field number compatibility broken across multiple regenerations: deleted field numbers are not marked as `reserved` and may be reused
### Background

[PR #357](https://github.com/tableauio/tableau/pull/357) introduced the `PreserveFieldNumbers` option to keep proto field numbers stable when regenerating proto files from sheets, in order to maintain binary (wire-format) compatibility of generated messages.

The current implementation in `sheetExporter.assignFieldNumbers` (`internal/protogen/exporter.go`) preserves field numbers by:

1. Reading existing field numbers from the previously generated `MessageDescriptor` (`oldMD`).
2. Tracking `maxFieldNumber` over fields currently present in `oldMD`.
3. Assigning `maxFieldNumber + 1` to any newly added field.

```401:431:internal/protogen/exporter.go
func (*sheetExporter) assignFieldNumbers(fields []*internalpb.Field, oldMD protoreflect.MessageDescriptor) {
    if oldMD == nil {
        // ...
    }
    fieldNameNumberMap := make(map[string]int32)
    var maxFieldNumber int32
    for i := 0; i < oldMD.Fields().Len(); i++ {
        fd := oldMD.Fields().Get(i)
        // ...
        maxFieldNumber = max(maxFieldNumber, int32(fd.Number()))
    }
    for _, field := range fields {
        if number, ok := fieldNameNumberMap[field.Name]; ok {
            field.Number = number
        } else {
            field.Number = maxFieldNumber + 1
            maxFieldNumber = field.Number
        }
    }
}
```

### Problem

PR #357 only protects field-number stability **within a single regeneration step**. Because deleted field numbers are **not emitted as `reserved`** in the generated proto, the information about deleted numbers is lost as soon as the proto file is written. On a subsequent regeneration, `oldMD` no longer contains those deleted fields, so `maxFieldNumber` shrinks and a previously deleted field number can be silently reused, breaking binary compatibility.

### Reproduction

Suppose a message originally has three fields:

```proto
message Foo {
  int32 a = 1;
  int32 b = 2;
  int32 c = 3;
}
```

**Case A — single regeneration step (works correctly):**

In one edit, delete field `c` and add field `d`. `oldMD` still contains fields 1, 2, 3, so `maxFieldNumber = 3` and `d` is assigned `4`. Result:

```proto
message Foo {
  int32 a = 1;
  int32 b = 2;
  int32 d = 4;
}
```

**Case B — two regeneration steps (broken):**

1. First edit: delete field `c`. After regeneration:
   ```proto
   message Foo {
     int32 a = 1;
     int32 b = 2;
   }
   ```
   Field number `3` is gone from the proto and is **not** marked `reserved`.

2. Second edit: add field `d`. On regeneration, `oldMD` only has fields 1 and 2, so `maxFieldNumber = 2` and `d` is assigned `3`:
   ```proto
   message Foo {
     int32 a = 1;
     int32 b = 2;
     int32 d = 3;
   }
   ```

Field number `3`, which previously belonged to `c`, is now reused by `d`. Any historical data serialized against the old schema will be incorrectly decoded as `d`, defeating the goal of PR #357.

The same issue applies to enum values and to nested/sub-messages that go through `assignFieldNumbers`.

### Proposed Fix

When a field present in `oldMD` is no longer in the new field list, its field number (and ideally its name) should be emitted as a `reserved` statement in the generated proto. The reserved set must also be carried forward across regenerations by reading existing `reserved` ranges/names from `oldMD` (`MessageDescriptor.ReservedRanges()` / `ReservedNames()`) and merging them with newly-deleted field numbers/names.

Concretely:

1. In `assignFieldNumbers`, collect the set of field numbers / names that exist in `oldMD` but not in the new `fields` slice.
2. Union that set with the existing reserved ranges/names already declared in `oldMD`.
3. Have the exporter emit something like:
   ```proto
   reserved 3, 7 to 9;
   reserved "c", "old_field";
   ```
   inside the generated message body.
4. Apply the same logic to enums (`reserved` for deleted enum numbers/names).

### Acceptance Criteria

- Deleting a field in one regeneration and adding a new field in a later regeneration never reuses the deleted field number.
- Existing `reserved` ranges/names in user-edited or previously generated protos are preserved across regenerations.
- Unit tests cover the multi-step delete-then-add scenario for messages, sub-messages, and enums.

Kybxd:
- close #402 

### Background / Motivation

When Tableau regenerates `.proto` files from Excel workbooks, it previously re-assigned field numbers strictly from 1..N in declaration order. This broke wire-format compatibility in two ways:

1. **Reordering / inserting columns** silently reassigned field numbers of unchanged columns, breaking existing serialized data.
2. **Deleting columns** freed the corresponding numbers for reuse by future new columns, violating protobuf's golden rule "never reuse a deleted field number".

This PR makes regeneration **wire-compatible** by preserving old field numbers and emitting `reserved` statements as needed.

### What changed

`internal/protogen/exporter.go` — `sheetExporter.assignFieldNumbers`:

- **Stable numbers for surviving fields.** When an `oldMD` (previously generated `MessageDescriptor`) is available, fields whose `name` matches an existing field keep their old number, regardless of column order.
- **Reserve deleted fields.** Any field present in `oldMD` but no longer in the current sheet has its number added to a `reserved` set.
- **Carry over pre-existing reserved ranges.** Reserved ranges already declared in `oldMD` are preserved verbatim across regenerations.
- **Auto-reserve gaps.** Numbers in `[1, maxUsedNumber]` that are neither active nor already reserved are auto-reserved — this defends against older Tableau versions that deleted fields without reserving them.
- **Collision-free new numbers.** New fields get the next number strictly greater than any active-or-reserved number, so they never collide with a reserved entry.

Supporting pieces:

- New helper `formatReservedNumbers` collapses a sorted number list into the compact proto syntax (`2, 5, 7 to 9, 12`).
- New helper `sheetExporter.printReserved` emits a single `reserved …;` line at the correct indentation.
- The three call-sites (`exportStruct`, `exportUnion`, `exportMessager`, and the nested-message branch in `exportField`) now consume the returned reserved slice and print it.

### Tests

`internal/protogen/exporter_test.go`:

- New `Test_formatReservedNumbers` covers empty / single / non-consecutive / consecutive / mixed / multi-range formatting.
- New `Test_sheetExporter_assignFieldNumbers` uses `protodesc.NewFile` to synthesize `MessageDescriptor`s with arbitrary active fields and reserved ranges, covering 10 semantic scenarios:
  - `oldMD == nil` → sequential assignment from 1.
  - All old fields kept → reuse old numbers.
  - New field appended → next number after max used.
  - Deleted field → number becomes reserved.
  - Pre-existing `reserved N` right after active fields → new field skips it.
  - Pre-existing `reserved N` higher than all active fields → all gaps below are auto-reserved; new field starts after N.
  - Pre-existing reserved + deletion in the same round → merged correctly.
  - Pre-existing consecutive `reserved 2 to 4` → carried over verbatim.
  - Active-field gap (1, 2, 4) without pre-existing reserved → gap auto-reserved.
  - Field rename → old number reserved, new name gets next number.
- Existing `exportStruct` / `exportUnion` / `exportMessager` tests extended with `field-number-compatibility-*` sub-tests (insertion-in-the-middle, delete-and-add, sub-message nesting).

### Compatibility

- Wire-format compatible: no existing field number is ever moved or reused.
- Backward compatible with `.proto` files generated by older Tableau versions that left gaps without `reserved`: those gaps are now auto-reserved on the next regeneration.
- No public API change; behavior is gated on whether `oldMD` is discoverable (i.e., requires `PreserveFieldNumbers` + an existing generated proto in the registry), so fresh generations are unaffected.

### DIFF
--- a/internal/protogen/exporter.go
+++ b/internal/protogen/exporter.go
@@ -1,6 +1,7 @@
 package protogen
 
 import (
+	"fmt"
 	"path/filepath"
 	"regexp"
 	"slices"
@@ -247,7 +248,8 @@ func (x *sheetExporter) exportStruct() error {
 	x.p.P("")
 
 	oldMD := x.findMDFromGeneratedProtos(x.ws.Name)
-	x.assignFieldNumbers(x.ws.Fields, oldMD)
+	reserved := x.assignFieldNumbers(x.ws.Fields, oldMD)
+	x.printReserved(1, reserved)
 	// generate the fields
 	depth := 1
 	for _, field := range x.ws.Fields {
@@ -358,7 +360,8 @@ func (x *sheetExporter) exportUnion() error {
 		if parentMD != nil {
 			oldMD = parentMD.Messages().ByName(protoreflect.Name(typ))
 		}
-		x.assignFieldNumbers(msgField.Fields, oldMD)
+		reserved := x.assignFieldNumbers(msgField.Fields, oldMD)
+		x.printReserved(depth, reserved)
 		for _, field := range msgField.Fields {
 			var oldFD protoreflect.FieldDescriptor
 			if oldMD != nil {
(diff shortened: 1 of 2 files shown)


wenchy (approved):
LGTM

PR #405, opened 2026-05-14
refactor(strcase,BREAKING!): adopt STYLE2024 case conversion rules

Kybxd:
## Motivation

The `internal/strcase` package was forked long ago from `iancoleman/strcase` and produced names that diverge from the Protobuf [STYLE2024](https://protobuf.dev/programming-guides/style/) and [Edition 2024: features.enforce_naming_style](https://protobuf.dev/editions/features/#enforce_naming) guide:

- It inserted underscores at letter↔digit boundaries (`Tier1` → `tier_1`), which STYLE2024 explicitly forbids.
- It treated all-caps acronyms as a single token (`JSONData` → various forms depending on the acronym table) instead of as ordinary words.
- It carried a long tail of unused helpers (`ToLowerCamel`, `ToKebab`, `ToDelimited`, …) inherited from upstream.

Meanwhile downstream code (`protogen`, `confgen`) kept patching around these rough edges with ad-hoc prefix logic and `ToCamel`/`ToPascal` swaps. This MR reworks the package to STYLE2024 in one place, removes the workarounds at call sites, and regenerates the affected test fixtures.

### `internal/strcase` — core rewrite

- New chunk-based PascalCase / snake_case algorithm in `camel.go` / `snake.go`. STYLE2024 invariants:
  - Underscores only appear before a letter (`Tier1` → `tier1`, never `tier_1`).
  - Acronyms are ordinary words (`GetDnsRequest` / `dns_request`, `userID` → `user_id`).
- Public API trimmed to the four conversions actually needed:

  | Function              | Result               |
  | --------------------- | -------------------- |
  | `ToPascal(s)`         | `AnyKindOfString`    |
  | `ToCamel(s)`          | `anyKindOfString`    |
  | `ToSnake(s)`          | `any_kind_of_string` |
  | `ToScreamingSnake(s)` | `ANY_KIND_OF_STRING` |

  `ToCamel` now produces `anyKindOfString` (i.e. previous `ToLowerCamel` semantics). `ToLowerCamel`, `ToKebab`, `ToScreamingKebab`, `ToDelimited`, `ToScreamingDelimited`, `ToSnakeWithIgnore` are removed.
- New helper `EnumValue(enumName, value)` (`enum_value.go`) producing STYLE2024-compliant enum value names: prefixes with `UPPER_SNAKE_CASE(enumName)`, normalizes the value, and inserts a leading `V` when the value would otherwise start with a digit. Examples:

  ```
  EnumValue("DeviceTier", "Tier1")           -> "DEVICE_TIER_TIER1"
  EnumValue("DeviceTier", "1")               -> "DEVICE_TIER_V1"
  EnumValue("ItemType",   "EQUIP")           -> "ITEM_TYPE_EQUIP"
  EnumValue("ItemType",   "ITEM_TYPE_EQUIP") -> "ITEM_TYPE_EQUIP"
  ```
- `doc.go` / `README.md` rewritten to describe the STYLE2024 contract and the (now smaller) public surface.
- Tests rewritten against the new rules; new `enum_value_test.go` covers `EnumValue` plus an acronym-context regression case.

### Call sites updated to use the new API

- `internal/confgen` (`util.go`, `table_parser.go`, `document_parser.go`): default field/column/node name derivation switches from `ToCamel` to `ToPascal` so the proto field `hero_id` continues to map to the sheet/document name `HeroId` under the new rules.
- `internal/protogen/exporter.go`: union enum value names go through `EnumValue("Type", field.Name)`; when an explicit message type is not provided, the type name is derived via `ToPascal(field.Name)`.
- `internal/protogen/sheet_mode.go`: enum value naming for sheet-mode enums goes through `EnumValue(ws.Name, value.Name)`; the ad-hoc prefix-prepend logic is removed.
- `internal/tools/cmd/ecode/main.go`: directly uses `ToCamel` for the Go field name (replaces the previous PascalCase + manual lower-first hack).

### Options — deprecation

`options.ProtoOutputOption.EnumValueWithPrefix` is **deprecated** and now a no-op. STYLE2024 mandates the enum-name prefix unconditionally, so the option no longer makes sense. The field is kept for config-schema backward compatibility; updated godoc reflects this. The functional test setup (`test/functest/util.go`) drops its usage.

## Breaking changes

This MR will change generated `.proto` content for any sheet whose names contain digit-letter junctions, all-caps acronyms, or rely on the ad-hoc enum prefix toggle. Concretely:

- Sheet/file names with digit junctions normalize differently:
  - `csv__bom__utf_8_bom.proto` → `csv__bom__utf8_bom.proto`
  - `excel__metasheet__merger_1.proto` → `excel__metasheet__merger1.proto`
  - `excel__metasheet__scatter_1.proto` → `excel__metasheet__scatter1.proto`
- Field / enum value spellings inside generated protos shift to the STYLE2024 form (see updated golden files under `test/functest/proto/default/*`).
- `EnumValueWithPrefix = false` no longer suppresses the prefix; downstream users relying on the unprefixed form must regenerate.
- Removed `strcase` symbols (`ToLowerCamel`, `ToKebab`, `ToDelimited`, …): no internal user remains, but external embedders of `internal/strcase` (if any) need to migrate to the four supported conversions.


wenchy:
Due to [Edition 2024: features.enforce_naming_style](https://protobuf.dev/editions/features/#enforce_naming)

> Values available:
> 
> STYLE2024: Enforces strict adherence to the style guide for naming.
> STYLE_LEGACY: Applies the pre-Edition 2024 level of style guide enforcement.

So if tableau support the new naming style, it shoud be enabled by config options.

PR #406, opened 2026-05-25
feat(load): support scatter sheets and configurable error capacity in `loadOrigin`

Kybxd:
## Motivation

`load.loadOrigin` (used by `LoadMessager` / `LoadMessagerInDir` to parse the *origin* workbook directly at runtime, bypassing pre-generated JSON/text) had two long-standing limitations:

1. It supported the **merger** sheet option but not **scatter**, even though `confgen` has supported scatter for offline generation for a while. As a result, any messager declaring `(tableau.worksheet).scatter = "..."` could not be loaded via the origin path.
2. Its error collector was hard-coded to capacity `1` (fail-fast). Callers wanting to surface multiple field-level errors in one shot — e.g. content validation tools, IDE-style integrations — had no knob to widen it.

This PR adds scatter support (parity with `confgen`'s semantics) and exposes the per-sheet error-collector capacity as a load option.

### `load/load.go` — scatter support in `loadOrigin`

- **Mutual exclusion with merger.** When `sheetOpts.GetScatter()` is non-empty, the new `loadOriginWithScatter` branch is taken; otherwise the existing merger branch runs. This matches `confgen`'s precedence rule.
- **Patch-mode dispatch** mirrors the offline pipeline:

  | `Patch` mode    | Behavior                                                                                      |
  | --------------- | --------------------------------------------------------------------------------------------- |
  | `PATCH_NONE`    | Only the main workbook is parsed; scatter shards are ignored.                                 |
  | `PATCH_REPLACE` | The last resolved scatter importer replaces the main message; falls back to main if no shard. |
  | `PATCH_MERGE`   | Main is parsed first, then each scatter shard is patched onto it via `xproto.PatchMessage`.   |

- The merger branch is refactored slightly so that "append self" happens explicitly only in that branch, keeping each path self-contained.
- A documentation comment is added on `loadOrigin` summarizing the new precedence + patch-mode contract.

### `load/options.go` — configurable error-collector capacity

- New field `BaseOptions.MaxErrorsPerSheet` (inherited by `MessagerOptions`) caps how many errors the collector aggregates while loading a single messager:
  - `<= 0` — fail-fast (behaves as `1`, current behavior). **Default.**
  - `1`    — fail-fast.
  - `> 1`  — aggregate up to N errors.
- `MessagerOptions.GetMaxErrorsPerSheet()` returns the effective value with the default already resolved.
- `Options.ParseMessagerOptionsByName` propagates the top-level value into per-messager options when the latter is unset, so users can configure once at `Options` level and override per messager when needed.
- `loadOrigin` now passes `opts.GetMaxErrorsPerSheet()` to `xerrors.NewCollector` instead of the hard-coded `1`.

> NOTE: this option only takes effect on the *origin* loading path (Excel/CSV/XML/YAML). Pre-generated JSON/text loading is unaffected.

### Tests & fixtures

- `load/load_test.go` — new `TestLoadOriginScatter` with three subtests, one per patch mode. Each subtest uses `LoadMessagerInDir` to drive the full origin pipeline end-to-end and asserts the merged proto with `proto.Equal`.
  - `PATCH_REPLACE` deliberately uses **a single** scatter shard: `importer.GetScatterImporters` returns importers in map-iteration order, so a multi-shard "last-wins" assertion would be flaky. This is documented inline.
- `proto/tableau/protobuf/unittest/unittest.proto` — three new test messages (`ScatterNoneConf`, `ScatterReplaceConf`, `ScatterMergeConf`) covering each patch mode. Each declares a `scatter:"UnittestScatter*.csv#<sheet>"` selector against the co-located fixtures.
- `testdata/unittest/*.csv` — fixtures for the three messages:
  - `Unittest#Scatter*Conf.csv`        — main workbook content.
  - `UnittestScatter1#Scatter*Conf.csv`, `UnittestScatter2#Scatter*Conf.csv` — scatter shards.
- `proto/tableaupb/unittestpb/unittest.pb.go` — regenerated.

## Backward compatibility

- **Default behavior is unchanged.** `MaxErrorsPerSheet` defaults to `0`, which resolves to `1` (the previous fail-fast behavior).
- Adding the scatter branch is strictly additive: messagers without `scatter` still go through the merger branch with identical behavior.
- No public API removed; `BaseOptions` / `MessagerOptions` only gain a new optional field.

### DIFF
--- a/load/load.go
+++ b/load/load.go
@@ -152,6 +152,16 @@ func Unmarshal(content []byte, msg proto.Message, path string, fmt format.Format
 
 // loadOrigin loads the origin file (excel/csv/xml/yaml) from the given
 // directory.
+//
+// Scatter and merger are mutually exclusive: scatter is processed first when
+// declared, otherwise the merger path is taken (consistent with confgen).
+//
+// For scatter sheets, the main workbook is always parsed first, and the
+// worksheet's patch mode determines how scatter importers are applied:
+//   - PATCH_NONE:    only the main workbook is loaded; scatter sheets are ignored.
+//   - PATCH_REPLACE: the last scatter importer replaces the main message.
+//   - PATCH_MERGE:   the main message is patched by each scatter importer
+//     in order via xproto.PatchMessage.
 func loadOrigin(msg proto.Message, dir string, opts *MessagerOptions) error {
 	md := msg.ProtoReflect().Descriptor()
 	protofile, bookOpts := confgen.ParseFileOptions(md.ParentFile())
@@ -178,14 +188,6 @@ func loadOrigin(msg proto.Message, dir string, opts *MessagerOptions) error {
 		return xerrors.Wrapf(err, "failed to import workbook: %v", wbPath)
 	}
 
-	// get merger importer infos
-	impInfos, err := importer.GetMergerImporters(context.Background(), dir, bookName, sheetName, sheetOpts.GetMerger(), subdirRewrites)
-	if err != nil {
-		return xerrors.Wrapf(err, "failed to get merger importer infos for %s", wbPath)
-	}
-	// append self
-	impInfos = append(impInfos, importer.ImporterInfo{Importer: self})
-
 	sheetInfo := &confgen.SheetInfo{
 		ProtoPackage:    string(md.ParentFile().Package()),
 		LocationName:    opts.GetLocationName(),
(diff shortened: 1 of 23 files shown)


load/load.go
```diff
@@ -200,7 +202,23 @@ func loadOrigin(msg proto.Message, dir string, opts *MessagerOptions) error {
 			BookFormat:     self.Format(),
 		},
 	}
-	collector := xerrors.NewCollector(1)
+	collector := xerrors.NewCollector(opts.GetMaxErrorsPerSheet())
+	mainImpInfo := importer.ImporterInfo{Importer: self}
+
+	// Scatter and merger are mutually exclusive: prefer scatter when declared.
+	if len(sheetOpts.GetScatter()) != 0 {
```
wenchy: `len(sheetOpts.GetScatter()) != 0` -> `sheetInfo.HasScatter()`

Refer
https://github.com/tableauio/tableau/blob/99d72fccd4a0bb70ddd363273b1161df196ec168/internal/confgen/parser.go#L247-L253

load/load.go
```diff
@@ -209,3 +227,72 @@ func loadOrigin(msg proto.Message, dir string, opts *MessagerOptions) error {
 	proto.Merge(msg, protomsg)
```
wenchy: Please refactor the scatter/merger/normal message parsing  dispatch logic just like the confgen. The goal is to improve readability and maintainance.

Refer
https://github.com/tableauio/tableau/blob/99d72fccd4a0bb70ddd363273b1161df196ec168/internal/confgen/confgen.go#L243-L263

load/load.go
```diff
@@ -209,3 +227,72 @@ func loadOrigin(msg proto.Message, dir string, opts *MessagerOptions) error {
...
+		if len(scatterImpInfos) != 0 {
+			impInfo = scatterImpInfos[len(scatterImpInfos)-1]
+		}
+		protomsg, err := confgen.ParseMessage(sheetInfo, collector, impInfo)
+		if err != nil {
+			return err
+		}
+		proto.Merge(msg, protomsg)
+
+	case tableaupb.Patch_PATCH_MERGE:
```
wenchy: Also treat the `patch == tableaupb.Patch_PATCH_NONE` as swith case dispatch for readability.

wenchy (changes requested):

wenchy (approved):
LGTM

wenchy (approved):

PR #421, opened 2026-06-23
ci: drop cmd/tableauc release tag scheme and bump buf-action

Kybxd:
## Background

Historically each release required pushing **two** tags pointing at the same commit:

- `vX.Y.Z` — the root module tag.
- `cmd/tableauc/vX.Y.Z` — required when `cmd/tableauc/` had its own `go.mod`.

The submodule `go.mod` was removed back in #181 (since v0.12.1), so `cmd/tableauc` is now just a regular sub-package of the root module `github.com/tableauio/tableau`. Users install it via `go install github.com/tableauio/tableau/cmd/tableauc@vX.Y.Z`, which resolves against the root tag — the `cmd/tableauc/vX.Y.Z` tag has had no effect on Go tooling since.

The only remaining consumer of `cmd/tableauc/*` was `release.yml`, where it was used as the workflow trigger and as part of the asset filename.

## Why this matters

Keeping the obsolete tag was actively harmful. When releasing v0.16.1, both tags were pushed to the same commit. Even though `buf-ci.yml` had filters specifically excluding `cmd/tableauc/*` from `push.tags`, the workflow still ran (triggered by the root `vX.Y.Z` tag), and `buf push --git-metadata` collects **all** refs pointing at HEAD. As a result `cmd/tableauc/v0.16.1` ended up published as a BSR label on `buf.build/tableauio/tableau`, polluting the module's label list.

### `release.yml`
- Trigger on top-level `vX.Y.Z` release tags instead of `cmd/tableauc/*`.
- Drop the leading `cd cmd/tableauc` from the `Download dependencies` step (no submodule `go.mod` to download against).
- Asset filename now derives from the root tag, e.g. `tableauc.v0.16.1.linux.amd64.tar.gz` (consistent with the historical naming style).
- Replace the deprecated `::set-output` syntax with `$GITHUB_OUTPUT`.

### `buf-ci.yml`
- Remove the `push.tags` allowlist that existed solely to keep `cmd/tableauc/*` out of buf-action runs — no longer needed.
- Add step names; unquote `version: 1.69.0` for consistency.

### `lint.yml`
- Bump `bufbuild/buf-action` to `1.69.0` to match `buf-ci.yml`.

## Release process going forward

Only the root tag is needed:

```bash
git tag v0.16.2
git push origin master v0.16.2
# Then publish a Release on GitHub against v0.16.2.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,9 +7,10 @@ on:
 
 jobs:
   release:
-    name: Release cmd/tableauc
+    name: Release tableauc
     runs-on: ubuntu-latest
-    if: startsWith(github.event.release.tag_name, 'cmd/tableauc/')
+    # Only run for top-level release tags like `vX.Y.Z` (optionally with a suffix).
+    if: startsWith(github.event.release.tag_name, 'v')
     strategy:
       matrix:
         goos: [linux, darwin, windows]
@@ -30,9 +31,7 @@ jobs:
           go-version: "1.24.x"
 
       - name: Download dependencies
-        run: |
-          cd cmd/tableauc
-          go mod download
+        run: go mod download
 
       - name: Prepare build directory
         run: |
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: .github/workflows/release.yml [hunk 1]
```diff
@@ -52,9 +51,11 @@ jobs:
       - name: Create package
         id: package
         run: |
-          PACKAGE_NAME=tableauc.${GITHUB_REF#refs/tags/cmd/tableauc/}.${{ matrix.goos }}.${{ matrix.goarch }}.tar.gz
+          # Strip the leading `refs/tags/` so that a tag like `v0.16.1`
+          # produces a package name `tableauc.v0.16.1.<goos>.<goarch>.tar.gz`.
+          PACKAGE_NAME=tableauc.${GITHUB_REF#refs/tags/}.${{ matrix.goos }}.${{ matrix.goarch }}.tar.gz
```
wenchy, line 56: DO NOT change the PACKAGE_NAME, keep it compatible and the cmd/tableauc path is good, mabye more cmds are developed in the future.

Revert it to: `GITHUB_REF#refs/tags/cmd/tableauc/`

Refer to https://github.com/tableauio/tableau/releases/tag/cmd%2Ftableauc%2Fv0.16.1

Image

wenchy (changes requested):

.github/workflows/release.yml
Kybxd: Unable to do this. `cmd%2Ftableauc%2FvX.Y.Z` is tag name not package name, but we just want to deprecate `cmd/tableauc/vX.Y.Z` tags in this pr.

wenchy (approved):

PR #429, opened 2026-07-06
feat: extract source comments as proto field notes for YAML and XML

Kybxd:
## Summary

Add the ability to extract comments from YAML and XML source documents and
emit them as `// ...` field comments in the generated proto files.

This introduces a `Note` field on `book.Node`, which is populated by the
importer during parsing and propagated to `internalpb.Field.Note` by the
document parser. The existing exporter then renders it as a trailing
field comment automatically — no exporter changes needed.

## YAML

Only trailing `# ...` line comments are collected (head/foot comments are
ignored for simplicity and to avoid ambiguous attribution).

```yaml
ID: uint32 # primary key
StructList:  # struct list
  "@type": "[Hero]"
```

The line comment may be attached to either the value node (when value is a
scalar on the same line) or the key node (when value spans multiple lines);
both are checked.

## XML

Two note syntaxes are supported in XML schema definitions, both using
attribute form with `@note` (consistent with `@type`):

1. **Element-level** — `@note="..."` attaches a note to the struct itself:

```xml
<Empty @note="empty placeholder struct">
</Empty>
```

This also works on text-only child elements:

```xml
<Date @note="creation date">date</Date>
```

2. **Field-level** — `@note.X="..."` attaches a note to the inline attribute
   field named X:

```xml
<Item ID="uint32" Name="string"
    @note.ID="item id"
    @note.Name="item name" />
```

Implementation: `parseXMLNode` pre-scans attributes to collect notes, then
skips note attributes during `parseXMLAttribute` calls and applies
collected notes to the corresponding field nodes via `applyFieldNote`.

`@note` uses the same displacement mechanism as `@type` (replaced to
`ANOTE` before XML parsing) since `@` is not valid in XML attribute names.

## Changes

- `book.Node`: add `Note string` field and `KeywordNote = "@note"` constant
- `yaml.go`: collect `LineComment` via `yamlLineNote`
- `xml.go`: collect `@note` and `@note.X` attributes (including element-level
  notes on text-only child elements and predefined structs with `@type`)
- `document_parser.go`: propagate `node.Note` to `field.Note`
- Tests: importer unit tests + functest golden protos for both YAML and XML
- All YAML and XML functest fixtures now have notes on every schema field

## Test plan

- [x] `go test ./internal/importer/...` — passes
  - `TestYAMLImporter_extractsCommentAsNote` (scalar/list/map/struct)
  - `TestXMLImporter_extractsNote` (element-level, inline attr, text-only
    child, nested struct field notes)
  - `TestXMLImporter_textOnlyChildNote` (text-only child with note, nested,
    special characters)
- [x] `go test ./internal/protogen/...` — passes
  - `TestDocumentParser_parseField_propagatesNote`
- [x] `go test ./test/functest/...` (Test_CompareGeneratedProto +
  Test_CompareGeneratedJSON) — passes
- [x] `go test ./...` — all green

## Limitations

- XML map key fields do not support notes (the key is stored as `@key` in
  the book node tree, so field-name lookup misses)
- YAML head/foot comments are intentionally ignored
- XML notes must use attribute form (`@note="..."` or `@note.X="..."`);
  `<!-- -->` comments are not supported


test/functest/testdata/default/xml/Struct.xml
```diff
@@ -6,16 +6,16 @@
 </@TABLEAU>
 
 <XMLStructConf>
-    <Item ID="uint32" Name="string" />
+    <Item ID="uint32" Name="string" note.ID="item id" note.Name="item name" />
     <Item2 ID="{OtherItem}uint32" Name="string" />
     <FruitShopData @type="{.FruitInfo}" />
-    <Empty>
+    <Empty note="empty placeholder struct">
```
wenchy: Need to add `@` prefix, just like `@type`.

`note="empty placeholder struct"` -> `@note="empty placeholder struct"`

Image

test/functest/testdata/default/xml/Struct.xml
```diff
@@ -6,16 +6,16 @@
...
     </Empty>
     <Award>{uint32 ID, int32 Num}Award</Award>
     <Vector>{.Vector3}</Vector>
-    <Date>date</Date>
-    <Text>string</Text>
-    <Attr Value="int32" />
+    <Date note="creation date">date</Date>
+    <Text note="display text">string</Text>
+    <Attr Value="int32" note.Value="attr value" />
```
wenchy: `note.Value="attr value"` -> `@note.Value="attr value"`

test/functest/testdata/default/yaml/List.yaml
```diff
@@ -41,7 +41,7 @@ MapInList:
 Items:
   "@type": "[Item]"
   "@struct":
-    ID: uint32|{unique:true}
+    ID: uint32|{unique:true} # item unique id
     Name: string|{unique:true}
     Num: int32
```
wenchy: Please add comments to all fieds. we wan to increase the test coverage.

Other YAML or XML should also be added comments to all fieds.

wenchy (changes requested):

test/functest/testdata/default/xml/Struct.xml
Kybxd: done

test/functest/testdata/default/xml/Struct.xml
Kybxd: done

test/functest/testdata/default/yaml/List.yaml
Kybxd: done

internal/importer/yaml.go
```diff
@@ -1,243 +1,274 @@
...
+	"io"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/tableauio/tableau/internal/importer/book"
+	"github.com/tableauio/tableau/internal/x/xerrors"
+	"github.com/tableauio/tableau/log"
+	"gopkg.in/yaml.v3"
```
wenchy, line 17: Why so many diffs? break line changes?

internal/importer/xml.go
```diff
@@ -1,721 +1,803 @@
...
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/subchen/go-xmldom"
+	"github.com/tableauio/tableau/internal/importer/book"
+	"github.com/tableauio/tableau/internal/importer/metasheet"
+	"github.com/tableauio/tableau/internal/types"
+	"github.com/tableauio/tableau/internal/x/xerrors"
+	"github.com/tableauio/tableau/log"
```
wenchy, line 20: Why so many diffs? break line changes?

wenchy (changes requested):

internal/importer/yaml.go
Kybxd: The whole-file diff is just line-ending normalization: the files on `master` use CRLF (`\r\n`), while the edited versions are LF (`\n`), so git marks every line as changed even though only the note additions are real. The actual content change in `yaml.go` is +31 lines (new `yamlLineNote` helper + wiring `Note:` for mapping/sequence). No unrelated logic was touched.

We're keeping LF as-is.

internal/importer/xml.go
Kybxd: Same as above — the whole-file diff is caused by CRLF→LF normalization (`master` is CRLF, edited files are LF), not by actual code changes. The real content change in `xml.go` is +100/-9: the `@note`/`@note.X` feature (displacement constants, pre-scan + `applyFieldNote`), relaxed predefined-struct check to allow meta attributes, and text-only child note support.

We're keeping LF as-is.

internal/importer/yaml.go
Kybxd: Use `Hide whitespace` to review real changes.
image

internal/importer/xml.go
```diff
@@ -1,721 +1,803 @@
...
+				continue
+			}
+			var err error
+			curBNode, err = parseXMLAttribute(curBNode, attr.Name, attr.Value, isFirstRealAttr)
+			if err != nil {
+				return xerrors.Wrapf(err, "parse xml attribute failed")
+			}
+			isFirstRealAttr = false
+			if note, ok := fieldNotes[attr.Name]; ok {
+				applyFieldNote(curBNode, attr.Name, note)
```
wenchy: **Bug: `@note.X` is silently dropped for vertical-map key attributes.**

`parseXMLAttribute`'s vertical-map branch (lines 481-498) builds the `@struct` container with a single child named `book.KeywordKey` (`"@key"`) whose **value** is `attrName`. `applyFieldNote` then walks `curBNode.Children` backwards looking for a child whose `Name == attrName` (e.g. `"Name"`), finds only the `@key` child, and returns without setting the note.

Reproducible from this PR's own testdata/golden:
- `test/functest/testdata/default/xml/map/Map.xml`: `<PreEntry Name="map<string, PreEntry>" @note.Name="pre entry name" Cost="[]int32" @note.Cost="pre entry cost" />`
- `test/functest/proto/default/xml__map__map.proto`: `cost_list` gets `// pre entry cost` but `string name = 1 [(tableau.field) = {name:"Name"}];` has **no** comment.

Same drop for `Meat.name`, `Vegetable.name`, `Fruit.name`. The test passes only because the golden was regenerated from the buggy code.

Suggested fix: have `parseXMLAttribute` return the created field node directly so the caller sets `.Note` without a name lookup (this also fixes the struct-non-first case below).

internal/importer/xml.go
```diff
@@ -1,721 +1,803 @@
...
+				},
+			}
+			if isFirstAttr {
+				bnode.Children = append(bnode.Children, &book.Node{
+					Name:  book.KeywordType,
+					Value: fmt.Sprintf("{%s}", desc.StructType),
+				}, curBNode)
+				return curBNode, nil
+			} else {
+				bnode.Children = append(bnode.Children, curBNode)
```
wenchy: **Bug: struct-typed attributes in non-first position drop their `@note.X`.**

This `default` (struct) branch appends `curBNode` itself — the `@struct` container named `"@struct"` — to `bnode` (line 635), unlike every other non-first branch (struct-list/scalar-list/incell-list at 523/558/594, which append `curBNode.Children[0]`, i.e. the field node named `attrName`). `applyFieldNote(bnode, attrName, note)` then searches `bnode.Children` for `attrName`, finds `"@struct"`, and silently no-ops.

Repro: `<Conf A="int32" B="{SomeStruct}" @note.B="b note"/>` -> `B`'s proto field is emitted without `// b note`. Inconsistent with sibling non-first branches that annotate correctly.

internal/importer/xml.go
```diff
@@ -1,721 +1,803 @@
...
+				Kind: book.MapNode,
+				Name: book.KeywordStruct,
+			}
+			bnode.Children = append(bnode.Children, &book.Node{
+				Name:  book.KeywordType,
+				Value: fmt.Sprintf("{%s}", node.Name),
+			}, curBNode)
+		}
+		bnode.Note = elemNote
+		for _, child := range node.Children {
```
wenchy, line 349: **Bug: `@note.X` targeting element-defined fields (child elements) is collected but never applied.**

The pre-scan (line 288) populates `fieldNotes[childName]` for any `@note.X`. But this children loop only applies a child's *own* `@note` attribute (`childNote`, lines 326-338) — it never consults the parent's `fieldNotes[child.Name]` for element children (lines 342-346).

Repro: `<Conf @note.Sub="sub note"><Sub>int32</Sub></Conf>` -> `fieldNotes["Sub"]` is set but `Sub`'s proto field gets no comment. The map is collected unconditionally, implying intent to support this; only attribute-defined fields are actually annotated.

internal/importer/yaml.go
```diff
@@ -1,243 +1,274 @@
...
+				ValuePos: book.Position{
+					Line:   value.Line,
+					Column: value.Column,
+				},
+				// Only trailing `# ...` line comments are collected. For a
+				// mapping pair the line comment may be attached to either
+				// the value node (when value is a scalar on the same line:
+				// `Key: value  # note`) or the key node (when value spans
+				// multiple lines: `Key:  # note\n  ...`).
+				Note: yamlLineNote(value.LineComment, key.LineComment),
```
wenchy: **Limitation: above-field YAML comments are dropped.**

`yamlLineNote` only reads `LineComment`. yaml.v3 attaches a comment on its own line above a field to `HeadComment`, leaving `LineComment` empty, so `# field note\nID: uint32` produces no comment on the generated `id` field.

This is documented as "trailing only" (line 168), but above-field comments are the most common commenting style in YAML schemas, so users will be surprised. Consider also checking `key.HeadComment`/`value.HeadComment` (with a preference order) if above-field notes should be supported.

internal/importer/xml.go
```diff
@@ -1,721 +1,803 @@
...
+		})
+		return bnode, nil
+	}
+}
+
+// applyFieldNote sets the note on the most recently created child named
+// fieldName of the struct container curBNode. It is used to attach a
+// `@note.X="..."` note to the corresponding field after parseXMLAttribute
+// has created it.
+func applyFieldNote(curBNode *book.Node, fieldName, note string) {
```
wenchy: **Simplification / root cause of the dropped-note bugs above.**

`applyFieldNote`'s backward-by-name child search is unnecessary: `parseXMLAttribute` contractually appends the new field as the **last child** of the returned container (its own comment says so). The name search (a) invents a false tolerance that silently no-ops on miss — hiding the vertical-map and struct-non-first cases instead of failing fast — and (b) is O(position) per noted field, O(n^2) across a struct where every field carries `@note.X`.

Prefer having `parseXMLAttribute` return the created field node directly so the caller sets `.Note` with no lookup. If a helper is kept, `curBNode.Children[len-1].Note = note` (with the existing nil guard) suffices.

internal/importer/xml.go
```diff
@@ -1,721 +1,803 @@
...
+			}
+			return nil
+		}
+		// NOTE: curBNode may be pointed to one subnode when needed
+		curBNode := bnode
+		// Pre-scan: collect element-level note and field-level notes
+		// (note.X="...") before processing real attributes, so that
+		// each field node can be annotated right after creation.
+		var elemNote string
+		fieldNotes := map[string]string{}
```
wenchy: **Efficiency: double attribute scan + unconditional per-node map allocation.**

This pre-scan allocates `fieldNotes := map[string]string{}` on every node and loops all attributes, then the main loop (292-305) re-runs `strings.HasPrefix` on every attribute. The map is allocated even when the sheet has zero `@note` attributes (the common case).

On a large XML schema sheet this doubles attribute iteration and allocates one map header per node. Cheaper: single-pass — dispatch non-note attrs immediately and attach the note to the just-created last child when a note attr is encountered (notes can be applied lazily since `parseXMLAttribute` always appends last).

internal/importer/yaml.go
```diff
@@ -1,243 +1,274 @@
...
+
+// yamlLineNote extracts a trailing `# ...` line comment from the first
+// non-empty candidate string. The leading "#" marker and surrounding
+// whitespace are stripped. Returns "" when no line comment is present.
+//
+// For a mapping pair the line comment may be attached to either the value
+// node (when value is a scalar on the same line: `Key: value  # note`)
+// or the key node (when value spans multiple lines: `Key:  # note`), so
+// both should be passed as candidates.
+func yamlLineNote(parts ...string) string {
```
wenchy: **Efficiency: variadic slice allocated per node.**

`yamlLineNote(parts ...string)` is called once per mapping pair and once per sequence element (lines 173, 200). The variadic signature constructs a `[]string` at each call site even though the common case (no comment) returns `""` after a single empty check.

For a large YAML workbook this is one slice per pair (heap-allocated if escape analysis can't prove non-escape). Cheaper: explicit args, e.g. `yamlLineNote(a, b string)` plus a single-arg form, or inline the empty-check at the call site and only call the helper when a comment is present.

internal/importer/xml.go
```diff
@@ -1,721 +1,803 @@
...
+	}
+	if err := scanner.Err(); err != nil {
+		log.Panicf("scanner err:%v", err)
+		return ""
+	}
+	metasheet := matches[1]
+	metasheet = strings.ReplaceAll(metasheet, "<@", "<AT")
+	metasheet = strings.ReplaceAll(metasheet, "</@", "</AT")
+	metasheet = strings.ReplaceAll(metasheet, book.KeywordType, atTypeDisplacement)
+	metasheet = strings.ReplaceAll(metasheet, book.KeywordNote, atNoteDisplacement)
```
wenchy: **Efficiency (minor): fourth full-string `ReplaceAll` over the metasheet.**

This adds a 4th `strings.ReplaceAll` over the entire metasheet comment content (on top of the 3 pre-existing calls at 711-713), each allocating a fresh full copy.

Bounded (once per book, not per node) so impact is small, but the four displacements can be combined into a single package-level `strings.Replacer` for one pass and one allocation.

internal/importer/book/node.go
```diff
@@ -42,6 +42,7 @@ const (
 	KeywordValue    = types.DefaultDocumentMapValueOptName // @value
 	KeywordIncell   = "@incell"
 	KeywordVariable = "@variable"
```
wenchy, line 44: **Altitude: `@note` is declared as a meta keyword but not wired into the shared meta machinery.**

`KeywordNote` sits alongside `@value`/`@incell`/`@variable`, but `@note` is only filtered ad-hoc inside `xml.go`'s attribute loops. It isn't integrated into `IsMeta()`/`GetChildrenWithoutMeta()`, and note extraction is duplicated in `xml.go` and `yaml.go` rather than living in `book/`.

Consequences: (1) any cross-cutting change to note semantics (trimming, multi-line joining, escaping, locale) must be implemented and tested in both importers and will drift; (2) a future `@`-meta keyword added to `book/node.go` won't be picked up by `xml.go`'s hand-maintained `xmlMetaAttrsAfterDisplacement` allowlist (line 52), so predefined-struct parsing would reject valid input until that map is also updated. Consider centralizing note population (and the meta-attr allowlist) in `book/`.

wenchy (commented):
## Code review — note extraction

Recall-biased review of the note-extraction feature. Findings ranked correctness first, then cleanup/efficiency/altitude.

**Correctness (notes silently dropped):**
1. `@note.X` dropped for **vertical-map key** attributes — *confirmed by this PR's own testdata/golden mismatch* (Map.xml `@note.Name="pre entry name"` -> golden `PreEntry.name` has no comment, while sibling `@note.Cost` works).
2. `@note.X` dropped for **struct-typed non-first** attributes (struct `default` branch appends the `@struct` container, not the field node — inconsistent with the list branches).
3. `@note.X` targeting **element-defined fields** is collected into `fieldNotes` but never applied in the children loop.
4. YAML **above-field comments** (`HeadComment`) are dropped — `yamlLineNote` reads only `LineComment`.

**Root cause / simplification:**
5. `applyFieldNote`'s backward-by-name child search is unnecessary and is the root cause of #1 and #2. `parseXMLAttribute` should return the created field node so the caller sets `.Note` directly.

**Efficiency:**
6. Double attribute scan + per-node `fieldNotes` map allocation in `parseXMLNode`.
7. Variadic `yamlLineNote(parts ...string)` allocates a slice per node.
8. (Minor) 4th `ReplaceAll` over the metasheet — fold into one `strings.Replacer`.

**Altitude:**
9. `KeywordNote` is declared as a meta keyword but not wired into `IsMeta()`/`GetChildrenWithoutMeta()`; note extraction is duplicated across `xml.go`/`yaml.go` rather than centralized in `book/`.

Details inline. The most actionable item is #1 — please confirm whether `// pre entry name` was intended on `PreEntry.name`; the testdata clearly expects it, which makes it a real defect rather than a golden drift.

wenchy (approved):

PR #435, opened 2026-07-09
perf: lazily compute proto registry; ci: fix Go setup and add generate check

Kybxd:
## Summary

- **CI: use `go-version-file: go.mod` for Go setup** — removes the hardcoded/matrix Go version in `testing.yml` and `release.yml`, so CI always builds with the exact Go version declared in `go.mod` instead of a value that can silently drift out of sync.
- **perf: lazily compute proto registry with generated files** — `ProtoRegistryFilesWithGenerated` is only needed by a few call paths (`PreserveFieldNumbers`, `preprocess(useGeneratedProtos=true, ...)`, the exporter's advanced mode). Previously it was eagerly parsed/compiled in every `NewGeneratorWithOptions` call regardless of whether it was used. This change turns it into a `sync.Once`-guarded lazy getter, avoiding the extra glob+compile IO/CPU cost on the common path while still computing it once and caching the result when actually needed.
- **CI: bump `golangci-lint-action` to `latest` and add a `go generate` drift check** — ensures generated code (e.g. `embed.go` outputs) is always committed and up to date, catching stale generated files at PR time instead of at review time.
- **CI: fix step ordering in `testing.yml`** — `Install Go` (which reads `go-version-file: go.mod`) was running before `Checkout Code`, so the repo wasn't checked out yet and `go.mod` couldn't be found. Reordered so checkout happens first.

## Verification

- `go build ./...`, `go vet ./...`, `go test ./...` all pass.
- `go generate ./...` is idempotent (no diff after running).

### DIFF
--- a/internal/protogen/protogen.go
+++ b/internal/protogen/protogen.go
@@ -47,12 +47,16 @@ type Generator struct {
 	InputOpt     *options.ProtoInputOption
 	OutputOpt    *options.ProtoOutputOption
 
-	ProtoRegistryFiles, ProtoRegistryFilesWithGenerated *protoregistry.Files
-	ProtoRegistryTypes                                  *dynamicpb.Types
+	ProtoRegistryFiles *protoregistry.Files
+	ProtoRegistryTypes *dynamicpb.Types
 
 	// internal
 	typeInfos *xproto.TypeInfos  // predefined type infos
 	collector *xerrors.Collector // concurrent error collector shared across the generator.
+
+	// used in advanced mode or when preserveFieldNumbers is set to true
+	registryWithGeneratedOnce       sync.Once
+	protoRegistryFilesWithGenerated *protoregistry.Files
 
 	cacheMu         sync.RWMutex                 // guard fields below
 	cachedImporters map[string]importer.Importer // absolute file path -> importer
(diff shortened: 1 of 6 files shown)


internal/protogen/protogen.go
```diff
@@ -86,15 +90,32 @@ func NewGeneratorWithOptions(protoPackage, indir, outdir string, opts *options.O
...
 	return gen
 }
 
+// GetProtoRegistryFilesWithGenerated returns a registry that includes both
+// the user-specified imported protos and the previously generated protos
+// under outdir. It parses on first call (under sync.Once) and caches the
+// result in [Generator.ProtoRegistryFilesWithGenerated]; subsequent calls
+// return the cached value. If the field is already set (e.g. by tests),
+// it is returned as-is without invoking the parser.
+func (gen *Generator) GetProtoRegistryFilesWithGenerated() *protoregistry.Files {
```
wenchy: No need to export this func as public API?

internal/protogen/protogen.go
```diff
@@ -117,7 +138,12 @@ func (gen *Generator) preprocess(useGeneratedProtos, delExisted bool) error {
 	// parse custom imported proto files
 	protoRegistryFiles := gen.ProtoRegistryFiles
 	if useGeneratedProtos {
-		protoRegistryFiles = gen.ProtoRegistryFilesWithGenerated
+		protoRegistryFiles = gen.GetProtoRegistryFilesWithGenerated()
+	}
+	if gen.OutputOpt.PreserveFieldNumbers {
+		// if preserveFieldNumbers enabled, parse generated protos before they
+		// are deleted
+		_ = gen.GetProtoRegistryFilesWithGenerated()
```
wenchy: 1. recursive call chain:  ` parseProtoRegistryFiles -> GetProtoRegistryFilesWithGenerated -> parseProtoRegistryFiles`
2. `if useGeneratedProtos || gen.OutputOpt.PreserveFieldNumbers` then `protoRegistryFiles = gen.GetProtoRegistryFilesWithGenerated()`

wenchy (changes requested):

internal/protogen/protogen.go
Kybxd: Good point, no need to export it. Renamed to `getProtoRegistryFilesWithGenerated` (unexported) in 79d66e7.

internal/protogen/protogen.go
Kybxd: 1. Good catch, that recursive call was a leftover from an earlier version - the current code already calls `parseProtoRegistryFiles` directly (not through `getProtoRegistryFilesWithGenerated`), so there's no recursion.
2. Merged the two `if`s into `if useGeneratedProtos || gen.OutputOpt.PreserveFieldNumbers { protoRegistryFiles = gen.getProtoRegistryFilesWithGenerated() }` as suggested, in 79d66e7.

wenchy (approved):

PR #437, opened 2026-07-10
feat(tableaupb): add Fraction.MatchAny/MatchAll, rename Cmp to Match

wenchy:
## Summary

Adds two multi-comparator helpers on `*Fraction` and renames the existing single-comparator method for a consistent naming family.

| API | Logic | Empty list |
|-----|-------|-----------|
| `Fraction.Match(cmp)` | single comparator (renamed from `Cmp`) | n/a |
| `Fraction.MatchAny(cmps...)` | OR | `false` (matches `slices.ContainsFunc`) |
| `Fraction.MatchAll(cmps...)` | AND | `true` (vacuous truth — identity element for AND) |

The package-level `Compare` is kept as a thin `// Deprecated:` wrapper over `Match` so external importers don't break, with a clear migration pointer.

## Why `Match` instead of `Cmp` / `Compare`

The singular method was renamed from `Cmp` → `Match`, and the new plural APIs use the same verb. The reasons:

1. **`Compare`/`Cmp` is a reserved Go idiom that returns an `int` ordering (−1/0/+1).** Across the standard library — `bytes.Compare`, `strings.Compare`, `bytes.(*Buffer).Cmp`, `cmp.Compare`, and the `slices.SortFunc` comparator — the name implies an integer ordering result. A `bool`-returning function named `Compare`/`Cmp` misleads readers into expecting an `int`, and obscures *why* `Cmp` is even usable as a `slices.ContainsFunc` predicate (it only type-checks because it returns `bool`).

2. **The operation is a predicate test, not a peer-to-peer comparison.** `Compare`/`Cmp` implies comparing two peers of the same kind. Here the operands are asymmetric: a `Fraction` (a value) tested against a `Comparator` (a constraint with a sign + threshold). The signature itself betrays this — you can't even "compare" two `Fraction`s, only test a `Fraction` against a `Comparator`.

3. **`Match` is an established Go `bool`-verb** (`regexp.MatchString`) that reads as a test at both the method (`f.Match(cmp)`) and package-func levels. It also unifies the singular and plural under one verb (`Match` / `MatchAny` / `MatchAll`), instead of the previous `Compare` (func) vs `Cmp` (method) split.

## Empty-list semantics

- `MatchAny` returns `false` for an empty list — consistent with `slices.ContainsFunc(nil, pred)` and the identity element for OR.
- `MatchAll` returns `true` for an empty list — vacuous truth, the identity element for AND (consistent with Python `all([])` and JS `[].every()`). This keeps the pair composable and De Morgan-consistent:
  ```
  MatchAll(left, append(cmps, c)) == MatchAll(left, cmps) && left.Match(c)   // holds iff empty == true
  MatchAny(left, append(cmps, c)) == MatchAny(left, cmps) || left.Match(c)   // holds iff empty == false
  ```

## Breaking change

`Fraction.Cmp` (method) and `tableaupb.Compare` (package func) are **exported on a published package**. `Compare` is preserved as a deprecated wrapper (no in-repo callers remain, so `staticcheck` SA1019 stays clean). `Cmp` is renamed outright — this is a breaking change for any external code calling `Fraction.Cmp`. Within this repo there are zero such callers (verified by grep).

## Verification

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./proto/tableaupb/` — `TestMatch`, `TestMatchAny`, `TestMatchAll` all pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/proto/tableaupb/wellknown_util.go
+++ b/proto/tableaupb/wellknown_util.go
@@ -1,5 +1,7 @@
 package tableaupb
 
+import "slices"
+
 // NewFraction creates a new fraction.
 func NewFraction(num, den int32) *Fraction {
 	return &Fraction{Num: num, Den: den}
@@ -29,11 +31,13 @@ func NewIntegerComparator(sign Comparator_Sign, num int32) *Comparator {
 }
 
 // Compare returns true if the given fraction matches the given comparator.
+//
+// Deprecated: Use [Fraction.Match] instead.
 func Compare(left *Fraction, cmp *Comparator) bool {
-	return left.Cmp(cmp)
+	return left.Match(cmp)
 }
 
-func (f *Fraction) Cmp(cmp *Comparator) bool {
+func (f *Fraction) Match(cmp *Comparator) bool {
 	other := cmp.GetValue()
 	// cross-multiply to compare
 	lval := int64(f.GetNum()) * int64(other.GetDen())
@@ -54,4 +58,21 @@ func (f *Fraction) Cmp(cmp *Comparator) bool {
 	default:
 		panic("invalid compare operator")
 	}
+}
+
+// MatchAny reports whether the fraction matches any of the given comparators (OR logic).
+// It returns false if no comparators are provided.
+func (f *Fraction) MatchAny(cmps ...*Comparator) bool {
+	return slices.ContainsFunc(cmps, f.Match)
+}
+
+// MatchAll reports whether the fraction matches all of the given comparators (AND logic).
+// It returns true if no comparators are provided (vacuous truth).
+func (f *Fraction) MatchAll(cmps ...*Comparator) bool {
+	for _, cmp := range cmps {
+		if !f.Match(cmp) {
+			return false
+		}
+	}
+	return true
 }
--- a/proto/tableaupb/wellknown_util.go
+++ b/proto/tableaupb/wellknown_util.go
@@ -37,9 +37,12 @@ func Compare(left *Fraction, cmp *Comparator) bool {
 	return left.Match(cmp)
 }
 
+// Match reports whether the fraction satisfies the comparator's
+// relational condition (==, !=, <, <=, >, >=). Comparison is done by
+// cross-multiplication to stay in exact integer arithmetic for performance
+// and accuracy.
 func (f *Fraction) Match(cmp *Comparator) bool {
 	other := cmp.GetValue()
-	// cross-multiply to compare
 	lval := int64(f.GetNum()) * int64(other.GetDen())
 	rval := int64(other.GetNum()) * int64(f.GetDen())
 	switch cmp.GetSign() {
@@ -56,7 +59,8 @@ func (f *Fraction) Match(cmp *Comparator) bool {
 	case Comparator_SIGN_GREATER_OR_EQUAL:
 		return lval >= rval
 	default:
-		panic("invalid compare operator")
+		// Unknown sign is treated as no-match; proto enums are open-ended (forward-compat), not a bug.
+		return false
 	}
 }
 
--- a/proto/tableaupb/wellknown_util_test.go
+++ b/proto/tableaupb/wellknown_util_test.go
@@ -71,6 +71,14 @@ func TestMatch(t *testing.T) {
 			},
 			want: true,
 		},
+		{
+			name: "unknown sign",
+			args: args{
+				left: NewFraction(1, 2),
+				cmp:  NewComparator(Comparator_Sign(42), 1, 2), // unrecognized sign → no match
+			},
+			want: false,
+		},
 	}
 	for _, tt := range tests {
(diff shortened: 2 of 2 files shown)


Kybxd (changes requested):
## 评审意见总结

### 主要改动要求（必须）
1. **`MatchAny` 改为循环实现**：当前用 `slices.ContainsFunc(cmps, f.Match)`，建议改成和 `MatchAll` 一致的显式 `for` 循环（见上方行内 `suggestion`）。这样两个方法风格统一、便于对比，同时能移除不再使用的 `slices` 导入（否则 `go build` 会失败）。

### 其他可优化点
2. **`Cmp`→`Match` 是破坏性变更**：你保留了 `Compare` 的 `// Deprecated:` 包装，但 `Fraction.Cmp` 被直接改名、没有任何兼容别名。该包已发布，`Fraction.Cmp` 的外部调用方会直接编译失败。建议二选一：
   - 保留一个 `// Deprecated: Use Match instead` 的 `Cmp` 别名（与 `Compare` 的处理方式一致）；或
   - 在 release notes 中明确标注这是破坏性变更。
   这是设计取舍，但对外发布的 API 值得考虑。

3. **（可选，非阻塞）nil 安全性**：`Compare(nil, cmp)` 与 `f.Match(cmp)` 在 `Fraction` 或 `cmp` 为 nil 时都会 panic。这是既有行为，但既然新增了 `MatchAny`/`MatchAll` 这类更“入口化”的 API，可考虑加一层 nil 守卫。可作为后续 follow-up。

4. **（可选，非阻塞）测试覆盖**：`TestMatch` 覆盖了 `Match` 但没有覆盖已废弃的 `Compare` 包装路径，加一个 trivial case 可防止 deprecated shim 静默失效；另外可以补一个 unknown `Comparator_Sign` 触发 panic 的测试，固化现有契约。

整体实现和测试质量都不错，主要就是第 1 点需要改，第 2 点请确认下破坏性变更的处理方式。

proto/tableaupb/wellknown_util.go
```diff
@@ -55,3 +59,20 @@ func (f *Fraction) Cmp(cmp *Comparator) bool {
 		panic("invalid compare operator")
 	}
 }
+
+// MatchAny reports whether the fraction matches any of the given comparators (OR logic).
+// It returns false if no comparators are provided.
+func (f *Fraction) MatchAny(cmps ...*Comparator) bool {
+	return slices.ContainsFunc(cmps, f.Match)
```
Kybxd, line 70: Please align `MatchAny` with `MatchAll` by using an explicit loop instead of `slices.ContainsFunc`. This keeps the two helpers stylistically consistent and lets us drop the now-unused `slices` import.

```suggestion
func (f *Fraction) MatchAny(cmps ...*Comparator) bool {
	for _, cmp := range cmps {
		if f.Match(cmp) {
			return true
		}
	}
	return false
}
```

Also remove `import "slices"`.

Kybxd (changes requested):
## Review Summary

### Required change
1. **`MatchAny` should use a loop**: Currently it uses `slices.ContainsFunc(cmps, f.Match)`. Please switch to an explicit `for` loop to match `MatchAll` (see the inline `suggestion`). This keeps the two helpers consistent and lets you remove the now-unused `slices` import (otherwise `go build` fails).

### Other optimization points
2. **`Cmp` → `Match` is a breaking change**: You kept a `// Deprecated:` wrapper for `Compare`, but `Fraction.Cmp` was renamed outright with no compatibility alias. Since this package is published, external callers of `Fraction.Cmp` will fail to compile. Suggest either:
   - keeping a `// Deprecated: Use Match instead` `Cmp` alias (consistent with how `Compare` is handled), or
   - clearly documenting the method rename as a breaking change in the release notes.
   This is a design trade-off, but worth considering for a published API.

3. **(Optional, non-blocking) nil safety**: `Compare(nil, cmp)` and `f.Match(cmp)` will panic when `Fraction` or `cmp` is nil. This is pre-existing behavior, but since `MatchAny`/`MatchAll` are more "entry-point" style APIs, a nil guard could make them safer. Could be a follow-up.

4. **(Optional, non-blocking) test coverage**: `TestMatch` covers `Match` but not the deprecated `Compare` wrapper path — adding a trivial case prevents the deprecated shim from silently rotting. Also consider a panic test for an unknown `Comparator_Sign` to lock in the existing contract.

Overall the implementation and tests are solid. Mainly #1 needs changing, and please confirm the handling of the breaking change in #2.

proto/tableaupb/wellknown_util.go
wenchy: But `slices.ContainsFunc` is prefered, see https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/modernize#hdr-Analyzer_slicescontains

Kybxd (approved):

PR #438, opened 2026-07-13
feat(load): protovalidate messages on the load path

Kybxd:
## What

Make the `load` (checker) path run protovalidate on fully-assembled messages, matching the coverage confgen already provides in its `storeMessage` step.

## Why

The `load` path is typically used by checkers and never reaches confgen's `storeMessage`, so protovalidate constraints (e.g. required fields, CEL rules) were previously not enforced when loading from input formats. This closes that gap.

## How

- `internal/confgen/util.go`: export `NewValidator` and `Validate` (previously unexported `validate`). `NewValidator` builds a validator whose extension type resolver is derived from `prFiles`, shared across the generate and load paths.
- `internal/confgen/confgen.go`: `GenAll` / `GenWorkbook` now call `NewValidator(prFiles)` instead of inlining `protovalidate.New(...)`.
- `load/load.go`: at the end of `loadOrigin`, validate the fully-assembled message via `confgen.Validate(msg, protovalidate.GlobalValidator)`. The global validator resolves extensions via `protoregistry.GlobalTypes`, which already contains custom predefined rules compiled into the binary, so no dedicated validator instance is needed on this path.

Validation is intentionally placed after the message is fully assembled (not inside `ParseMessage`), to avoid false positives on intermediate patch fragments and duplicate validation.

## Test

`go build ./...` and `go test ./load/... ./internal/confgen/...` pass.

### DIFF
--- a/load/load.go
+++ b/load/load.go
@@ -25,6 +25,16 @@ import (
 
 // LoadMessagerInDir loads message's content in the given dir, based on format and messager options.
 func LoadMessagerInDir(msg proto.Message, dir string, fmt format.Format, opts *MessagerOptions) error {
+	if err := loadMessagerInDir(msg, dir, fmt, opts); err != nil {
+		return err
+	}
+	// protovalidate the fully-assembled message. protovalidate operates on the
+	// final in-memory protobuf, so both input (excel/csv/xml/yaml) and output
+	// (json/binpb/txtpb) formats get the same coverage as confgen's storeMessage.
+	return confgen.Validate(msg, protovalidate.GlobalValidator)
+}
+
+func loadMessagerInDir(msg proto.Message, dir string, fmt format.Format, opts *MessagerOptions) error {
 	if format.IsInputFormat(fmt) {
 		return loadOrigin(msg, dir, opts)
 	}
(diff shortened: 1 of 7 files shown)


load/load.go
```diff
@@ -202,12 +203,18 @@ func loadOrigin(msg proto.Message, dir string, opts *MessagerOptions) error {
...
 	default:
-		return loadOriginMain(msg, sheetInfo, mainImpInfo, collector)
+		err = loadOriginMain(msg, sheetInfo, mainImpInfo, collector)
+	}
+	if err != nil {
+		return err
 	}
+	// protovalidate the fully-assembled message, giving the load path the same
+	// coverage as confgen's storeMessage step.
+	return confgen.Validate(msg, protovalidate.GlobalValidator)
```
wenchy: We'd better add test case to cover the error condition of confgen.Validate, expect protovalidate error. This accounts for future regression.

> Regression testing is typically conducted after bug fixes are implemented, and it involves re-running tests to ensure features function as expected after new code is added.

wenchy (changes requested):

load/load.go
Kybxd: Done. Added a regression test in `load/load_test.go` (`TestLoadOriginProtovalidate`) plus a new `testdata/unittest/Unittest#ValidateConf.csv` fixture whose `Name` exceeds `max_len:10`. Loading it through the full load path asserts the violation surfaces as `xerrors.ErrE2027`, so the test will fail if the `confgen.Validate` call in `loadOrigin` is ever dropped.

Actual output:
```
error[E2027]: protovalidate violation
Reason: "this name is too long" violates rule: name: must be at most 10 characters
```

wenchy (approved):

load/load.go
wenchy: protovalidate should work both on  input (origin formats: xlsx/csv/yaml/xml) and output file formats(json/binpb/txtpb), as protovalide works on the final in memory protobuf message structure.

wenchy (changes requested):

load/load.go
Kybxd: Good point. Moved the `confgen.Validate` step out of `loadOrigin` and converged it at the single `LoadMessagerInDir` exit, right after the message is fully assembled (and after patch merging, to avoid false positives on intermediate state). Now both input (xlsx/csv/yaml/xml) and output (json/binpb/txtpb) formats validate the final in-memory protobuf message, matching confgen's `storeMessage` behavior.

Extended the regression test into `TestLoadProtovalidate` with two subcases and added a JSON fixture:
- `input-format-csv` — loads `ValidateConf` via CSV
- `output-format-json` — loads `testdata/unittest/conf/ValidateConf.json`

Both assert the violation surfaces as `xerrors.ErrE2027`:
```
error[E2027]: protovalidate violation
Reason: "this name is too long" violates rule: name: must be at most 10 characters
```

wenchy (approved):

PR #444, opened 2026-08-04
feat(protogen): shard huge union types via WorksheetOptions.union_shard_size

hazy-wu:
## Summary

Add an opt-in, per-sheet setting for sharding large `MODE_UNION_TYPE` and `MODE_UNION_TYPE_MULTI` sheets into multiple proto files.

The `@TABLEAU` metasheet column `UnionShardSize` maps to `WorksheetOptions.union_shard_size`. A value of `0` keeps the existing nested union output. A positive value emits message-valued union variants into shard files containing at most that many messages each.

## Motivation

Large unions can produce multi-megabyte generated translation units and slow downstream builds. Sharding the affected sheet reduces the size of the largest translation unit and allows shard files to compile in parallel, while leaving ordinary unions unchanged.

## Design

- Add `WorksheetOptions.union_shard_size = 25`; `0` disables sharding.
- Add the matching `UnionShardSize` metasheet column at field `26`.
- Emit extracted messages as top-level types named `<Union><SubType>`.
- Keep scalar, enum, predefined, and reused union references valid; shard files collect the imports required by their fields, including predefined types from `common.proto`.
- Store generated files as `unionShardFiles` and keep each shard's `ImportPath` with its generated content and imports.
- Derive shard filenames from the main proto basename, sheet name, shard index, and existing filename suffix.
- Check the main proto and all shard paths for conflicts before writing any output.
- Preserve field numbers and reserved ranges by looking up existing top-level shard descriptors first, then nested union descriptors.
- Omit `union_shard_size` from generated `option (tableau.worksheet)` output because it controls generation rather than the runtime schema.

## Behavior

| Configuration | Result |
|---|---|
| `union_shard_size = 0` (default) | Existing nested union output |
| `union_shard_size > 0` | Message-valued variants are emitted in imported shard files |

## Tests

- Unit coverage for default behavior, one-message shards, multiple shards, mixed scalar/predefined/reused types, custom names, filename suffixes, and field-number preservation.
- Functional coverage for a shard file importing `common/common.proto` and referencing both predefined `FruitType` and `Item` types.
- Existing generated fixtures remain consistent with the updated behavior.

Validation:

- `go test ./internal/protogen/...`
- `go test -run '^Test_CompareGeneratedProto$' -count=1` from `test/functest`


proto/tableau/protobuf/tableau.proto
```diff
@@ -213,6 +213,32 @@ message WorksheetOptions {
...
+  // sub-messages nested inline as usual.
+  //
+  // Default: 0 (disabled).
+  int32 union_split_threshold = 25;
+  // Number of extracted sub-messages placed in each shard file when
+  // `union_split_threshold` triggers splitting. Only used when
+  // `union_split_threshold > 0`.
+  //
+  // Default: 40.
+  int32 union_split_shard_size = 26;
```
wenchy: No need to support globally.

proto/tableau/protobuf/internal/metabook.proto
```diff
@@ -164,6 +164,28 @@ message Metasheet {
...
+  }];
+  // Number of extracted sub-messages placed in each shard file when
+  // `union_split_threshold` triggers splitting. Only used when
+  // `union_split_threshold > 0`.
+  //
+  // Default: 40.
+  int32 union_split_shard_size = 27 [(tableau.field) = {
+    name: "UnionSplitShardSize"
+    prop: {optional: true}
+  }];
```
wenchy: Just one option is enough:
```
 // Number of oneof message defintions placed in each shard file for Union Type definition. 
  //
  // Default: 50.
  int32 union_shard_size = 26 [(tableau.field) = {
    name: "UnionShardSize"
    prop: {optional: true}
  }];
```

wenchy: DO NOT ADD verbose `T_`, use `<Union><SubType>` instead for clean code and protobuf conventions.
 `<Union>T_<SubType>` ->  `<Union><SubType>`

wenchy: Missing functest for protogen at https://github.com/tableauio/tableau/tree/master/test/functest

Add csv files for Union type definiton, and use the new option "UnionShardSize"

wenchy (changes requested):

proto/tableau/protobuf/tableau.proto
hazy-wu: To clarify "no globally": should UnionShardSize live only in Metasheet (with a new internal-only field on internal.Worksheet to reach the exporter), or is keeping it in WorksheetOptions acceptable?

Kybxd (changes requested):
Thanks for this PR! The overall direction looks good — the per-sheet split strategy is the right level of granularity. Below are the detailed review comments based on discussing with the team.

---

### 1. (Blocking) Merge two config fields into a single `union_shard_size`

**Files**: `proto/tableau/protobuf/tableau.proto`, `proto/tableau/protobuf/internal/metabook.proto`, `internal/protogen/exporter.go`

As wenchy pointed out, there is no need for two separate fields (`union_split_threshold` + `union_split_shard_size`). A single field is sufficient:

```proto
// Number of oneof sub-messages placed in each shard file for Union Type definition.
// Set to 0 to disable splitting (the default).
// Default: 0 (disabled).
int32 union_shard_size = 25;
```

Semantics:
- `union_shard_size = 0` (default) → no split, original inline behavior
- `union_shard_size > 0` → split enabled, each shard holds up to N sub-messages

This also simplifies the judgment logic in `exportUnion()`:
```go
shardSize := int(x.ws.GetOptions().GetUnionShardSize())
shouldSplit := shardSize > 0
```

### 2. (Blocking) Remove the `T_` separator — use flat `<Union><SubType>` directly

**File**: `internal/protogen/exporter.go`

We evaluated two alternatives:

**Option A (recommended) — flat concatenation**: `<Union><SubType>`, e.g. `TaskTargetPvpBattle`
**Option B — nested wrapper**: `<Union><ShardN>` enclosing `<SubType>`, e.g.
```protobuf
message TaskTarget1 {
  message PvpBattle { ... }
}
```

We recommend **Option A**. Option B adds unnecessary nesting, making the oneof field declarations more verbose (`TaskTarget1.PvpBattle pvp_battle = 1`) without solving a real problem — shard files already isolate namespaces via the file path.

Regarding the `T_` collision rationale from the current PR description: protoc-gen-go's oneof wrapper naming (`<Parent>_<CamelField>`) is an internal Go type name, not a proto-level name. It does not collide with proto-level `message` definitions across files, since protoc resolves types file-by-file via imports. Therefore `<Union><SubType>` is safe to use without `T_`.

Actions:
- Delete constant `unionSplitTypeSep`
- In `exportUnion()`, change oneof field type to `x.ws.Name + typ`
- In `exportUnionSplit()`, change `prefixedTyp` to `x.ws.Name + sm.typ`

### 3. (Blocking) Missing functest

**Directory**: `test/functest/`

A functest case is needed. Add a CSV file defining a Union type sheet with `UnionShardSize` set and verify the end-to-end proto generation output.

### 4. (Blocking) Remove `defaultUnionSplitShardSize` — split only when user explicitly sets a positive value

**File**: `internal/protogen/exporter.go`

The current code has `defaultUnionSplitShardSize = 40`, which means setting `union_split_threshold > 0` but leaving `union_split_shard_size = 0` still triggers splitting. Per discussion, the user's explicit opt-in should be required for every aspect: `union_shard_size = 0` means **no split at all**. Delete the `defaultUnionSplitShardSize` constant and the fallback logic.

### 5. (Blocking) Extract common proto file header generation to eliminate duplication

**File**: `internal/protogen/exporter.go`

`export()` and the new `writeAuxFile()` both duplicate ~30 lines of header generation code (the "Code generated by tableau" comment, syntax/edition, package, sorted imports, sorted file options). Extract this into a shared method:

```go
func (x *bookExporter) writeProtoHeader(p *printer.Printer, imports map[string]bool) {
    // comment, syntax/edition, package, sorted imports, sorted file options
}
```

Both `export()` and `writeAuxFile()` should call this method. `writeAuxFile()` then becomes much simpler: call `writeProtoHeader` → write the shard body.

### 6. (Suggestion) Add boundary test cases

**File**: `internal/protogen/exporter_test.go`

The current `export-union-split` test only covers `threshold=1, shardSize=1` with 2 sub-messages. Consider adding:
- `shardSize=0` → no split (default)
- `shardSize=1` with only 1 sub-message → single shard
- Mixed local types and external predefined types → verify external types are excluded from subMsgs

### 7. (Suggestion) Use `filepath.Ext` for shard file naming

**File**: `internal/protogen/exporter.go`, `exportUnionSplit()`

The current `strings.TrimSuffix(mainPath, x.be.FilenameSuffix+".proto")` is fragile when `FilenameSuffix` is empty or non-standard. Use `filepath.Ext` + `filepath.Base` instead:

```go
mainPath := x.be.GetProtoFilePath()
ext := filepath.Ext(mainPath)
base := strings.TrimSuffix(filepath.Base(mainPath), ext)
shardTag := strcase.FromContext(x.be.gen.ctx).ToSnake(x.ws.Name)
shardRelPath := fmt.Sprintf("%s_%s_%d%s", base, shardTag, shardIdx+1, ext)
```

Keep the `shardTag` — it is necessary to disambiguate when a single workbook proto contains multiple union sheets.

### 8. (Confirmed) No additional mode validation needed

No need to add a warning when `union_shard_size` is set on non-union sheets. Other parameters (e.g. `Index` only effective for `MODE_DEFAULT`) already follow this pattern of being silently ignored for inapplicable modes.

---

Summary of action items:
| # | Type | Summary |
|---|------|--------|
| 1 | Blocking | Merge into single `union_shard_size` field, `>0` enables, `=0` disables |
| 2 | Blocking | Use `<Union><SubType>` flat naming, remove `T_` separator |
| 3 | Blocking | Add functest |
| 4 | Blocking | Remove `defaultUnionSplitShardSize`, no implicit default |
| 5 | Blocking | Extract shared `writeProtoHeader` to eliminate header duplication |
| 6 | Suggestion | Add boundary test cases |
| 7 | Suggestion | Use `filepath.Ext` for shard file naming |
| 8 | Confirmed | No mode validation needed |

proto/tableau/protobuf/tableau.proto
Kybxd: I suggest in `internal.Worksheet` only.

proto/tableau/protobuf/internal/workbook.proto
```diff
@@ -21,6 +21,9 @@ message Worksheet {
   string note = 2;
   tableau.WorksheetOptions options = 3;
   repeated Field fields = 4;
+  // Number of oneof sub-messages placed in each shard file for Union Type
+  // definition. Set to 0 to disable splitting (the default).
+  int32 union_shard_size = 5;
```
wenchy: `union_shard_size`  already defined in  `WorksheetOptions`, and kept by this Worksheet message, no need to redefine it here.

internal/protogen/exporter.go
```diff
@@ -35,6 +35,21 @@ type bookExporter struct {
 	gen *Generator
 
 	messagerPatternRegexp *regexp.Regexp
+
+	// auxFiles holds auxiliary shard proto files produced by union splitting
+	// (see Worksheet.union_shard_size). Empty when no union is split.
+	auxFiles []*unionAuxFile
+	// auxImportPaths holds the relative paths of aux files that the main proto
+	// file needs to `import`. Populated together with auxFiles.
+	auxImportPaths []string
```
wenchy: Just let the `unionAuxFile` contains the `auxImportPath`

Do not use confusing aux, use full word:

aux -> auxiliary

wenchy (changes requested):

proto/tableau/protobuf/internal/workbook.proto
wenchy: The PR description and title should also be updated!

wenchy (approved):

PR #447, opened 2026-08-12
fix: resolve proto imports and Excel cleanup

wenchy:
## Summary
- resolve generated protos using the most specific matching proto root
- prevent duplicate symbol errors when proto roots overlap
- add regression coverage for overlapping roots and out-of-root file fallback
- close Excel row iterators before cleanup

## Test Plan
- go test ./...


internal/x/xproto/protoc/compiler.go
```diff
@@ -83,9 +107,9 @@ func parseProtos(protoPaths []string, protoFilesMap map[string]string) (*protore
 			&protocompile.SourceResolver{
 				ImportPaths: protoPaths,
 				Accessor: func(path string) (io.ReadCloser, error) {
-					if _, ok := protoFilesMap[xfs.CleanSlashPath(path)]; !ok {
-						return nil, fs.ErrNotExist
-					}
+					// protoFilesMap selects the compilation entry points. Their imports
+					// must still be resolved from the configured proto paths, even when
+					// an imported dependency is not matched by protoFiles.
```
Kybxd: This is designed deliberately to prevent predefined protos from importing generated protos. Do not remove.

Kybxd (approved):

PR #448, opened 2026-08-12
feat(protogen): support repeated field inside incell struct

Kybxd:
## What

Allow declaring a repeated element type inside an **incell struct** definition, e.g.

```
{int32 ID, string Name, []string Desc}Property
```

The declared name should be the **singular** form. The generated proto field name is automatically suffixed with `_list`, while the original name is preserved in `(tableau.field).name` for data-column matching (consistent with the existing header-list convention).

### Example

| Declared in cell | Generated proto |
|---|---|
| `[]int32 ID` | `repeated int32 id_list = N [(tableau.field) = {name:"ID"}]` |
| `[]string Desc` | `repeated string desc_list = N [(tableau.field) = {name:"Desc"}]` |
| `[]enum<.ItemType> Type` | `repeated ItemType type_list = N [(tableau.field) = {name:"Type"}]` |

### Data format

Repeated elements inside the cell are split by the sheet's `subsep`. For example, with `sep:"," subsep:":"`, a row value `"1,Apple,A kind of fruit.:Sweet and juicy."` produces `Desc = ["A kind of fruit.", "Sweet and juicy."]`.

## Why

Incell structs previously only supported scalar/enum fields. Real configs often need a small list nested inside a compact incell struct (tags, descriptions, aliases, ...). Before this change, users had to promote the whole struct into a vertical/horizontal layout just to hold one list, which is much more verbose.

## Changes

- `internal/protogen/parser.go`
  - New `parseIncellStructField` handles the `[]ElemType` form: only scalar / enum elements are allowed, nested composites (`[[`, `{`) are rejected.
  - Element field is built by reusing `parseBasicField`, then promoted to `repeated` by setting `ListEntry` and prefixing `Type` / `FullType`.
  - Generated field name is `ToSnake(singular_name) + "_list"`; `Options.Name` keeps the user's original spelling.
- `internal/protogen/table_parser.go`, `internal/protogen/document_parser.go`
  - Route incell struct subfield parsing through the new helper.
- `internal/types/types.go`
  - Allow `[` and `]` in `structTypeCharSet` so the type parser accepts `[]T` inside `{ ... }`.
- `internal/confgen/parser.go`
  - When parsing an incell struct, repeated subfields split the cell chunk by `subsep`.
- Tests
  - `internal/protogen/parser_test.go`: new cases for repeated scalar / enum incell fields; lexer cases updated.
  - `internal/types/types_test.go`: struct-type lexer cases for `[]T` element type.
  - `test/functest`: extended the existing `InCellStruct` sheet with a `[]string Desc` column; proto & JSON goldens updated. No new sheet is added.

## Compatibility

- Existing incell struct declarations are unchanged — this is purely additive.
- Users who want the repeated behavior must opt in by writing `[]T Name` in the type row.

## Test

- `go test ./internal/protogen/... ./internal/confgen/... ./internal/types/...` — passes.
- `./test/functest/run.sh` — passes with no golden diff.


internal/confgen/parser.go
```diff
@@ -866,6 +866,26 @@ func (p *sheetParser) parseIncellStruct(field *Field, structValue protoreflect.V
 				subField.mergeParentFieldProp(field)
 				defer subField.release()
 				// log.Debugf("fd.FullName().Name(): ", fd.FullName().Name())
+				if fd.IsList() {
+					// repeated sub-field: split by subsep, and require
+					// subsep to be different from the struct-field sep,
+					// otherwise the two levels cannot be distinguished.
+					if subField.subsep == "" || subField.subsep == sep {
+						return xerrors.Newf("repeated field in incell struct requires a distinct subsep (sep=%q, subsep=%q)", sep, subField.subsep)
+					}
```
wenchy: Seems verbose, not required due to previous guarantee.

internal/protogen/parser.go
```diff
@@ -149,23 +149,85 @@ func parseTypeDescriptor(typeInfos *xproto.TypeInfos, rawType string) (*types.De
...
+// It accepts scalar/enum types, and also a "repeated" form `[]ElemType` which
+// produces a repeated proto field. The generated field name is auto-suffixed
+// with "_list", so users should declare the singular form (e.g. `[]int32 ID`
+// generates `repeated int32 id_list`).
+func (p *bookParser) parseIncellStructField(name, typ, note string) (*internalpb.Field, error) {
+	return parseIncellStructField(p.gen.ctx, p.gen.typeInfos, name, typ, note)
+}
+
+// parseIncellStructField parses one field of an incell struct definition.
+// It accepts scalar/enum types, and also a "repeated" form `[]ElemType` which
```
wenchy: only scalar/enum types? what about wellknown types: google.protobuf.Timesamp..., which should also be supported.

NOTE: scalar contains wellknown types?

Add tests to cover wellknown types.

internal/protogen/parser.go
```diff
@@ -149,23 +149,85 @@ func parseTypeDescriptor(typeInfos *xproto.TypeInfos, rawType string) (*types.De
...
-	for _, pair := range strings.Split(structType, ",") {
-		kv := strings.Split(strings.TrimSpace(pair), " ")
-		if len(kv) != 2 {
+	fieldPairs := make([]string, 0, len(fields)*2)
+	for _, pair := range fields {
+		// A pair is "<type> <name>". Only split by the LAST space so that the
+		// type part can contain spaces (currently not used, but safer against
+		// future extension).
+		tokens := strings.Fields(strings.TrimSpace(pair))
+		if len(tokens) < 2 {
```
wenchy: > type part can contain spaces (currently not use)

Bad decision, keep rule strict as before: `if len(tokens) != 2`, then report an error.

internal/types/types_test.go
```diff
@@ -288,6 +288,27 @@ func TestMatchStruct(t *testing.T) {
...
+			},
+			want: &StructDescriptor{
+				StructType: "[]int32 ID, string Name",
+				ColumnType: "Property",
+			},
+		},
+		{
+			name: "new-defined-struct-with-multiple-repeated-fields",
+			args: args{
+				text: `{[]int32 ID, []string Name}Property|{range:"1,10"}`,
```
wenchy: Any meaning of field property `|{range:"1,10"}`  for incell struct?

test/functest/testdata/default/excel/struct/Struct#InCellStruct.csv
```diff
@@ -1,6 +1,6 @@
 ID,Prop
-"map<int32, Item>","{int32 ID,string Name,string Desc}Property"
+"map<int32, Item>","{int32 ID,string Name,[]string Desc}Property"
```
wenchy: Also add a `datetime` list field for wellknown type test.

wenchy (changes requested):

internal/confgen/parser.go
Kybxd: Good point, removed the redundant empty check. `subsep` is guaranteed non-empty upstream (falls back to `options.DefaultSubsep`), so now only the `subsep == sep` conflict check remains, with a comment explaining the guarantee.

internal/protogen/parser.go
Kybxd: You're right, well-known types (e.g. `google.protobuf.Timestamp`/`Duration`/...) are already classified as `types.ScalarKind` by `types.ParseTypeDescriptor`/`MatchScalar`, so they're supported by the existing `elemDesc.Kind != types.ScalarKind && elemDesc.Kind != types.EnumKind` check without any code change. Updated the doc comment and error message to explicitly mention well-known types, and added unit tests covering `[]datetime` and `[]duration` repeated fields.

internal/protogen/parser.go
Kybxd: Agreed, reverted to the strict `len(tokens) != 2` rule and dropped the "type part can contain spaces" handling.

internal/types/types_test.go
Kybxd: No, `ExtractStructFieldProp` discards this prop for incell struct fields, so it has no actual effect. Removed the meaningless `range` prop from the test case.

test/functest/testdata/default/excel/struct/Struct#InCellStruct.csv
Kybxd: Added a `[]datetime CreatedAts` list field to the `InCellStruct` case (using pure date values to avoid clashing with the default subsep `:`), and updated the corresponding golden proto/conf files.

test/functest/proto/default/excel__struct__struct.proto
```diff
@@ -41,11 +42,12 @@ message InCellStruct {
...
   message Item {
     int32 id = 1 [(tableau.field) = {name:"ID"}]; // Item ID
-    Property prop = 2 [(tableau.field) = {name:"Prop" span:SPAN_INNER_CELL}]; // Item property
+    Property prop = 2 [(tableau.field) = {name:"Prop" span:SPAN_INNER_CELL prop:{subsep:";"}}]; // Item property
     message Property {
       int32 id = 1 [(tableau.field) = {name:"ID"}];
       string name = 2 [(tableau.field) = {name:"Name"}];
-      string desc = 3 [(tableau.field) = {name:"Desc"}];
+      repeated string desc_list = 3 [(tableau.field) = {name:"Desc"}];
+      repeated google.protobuf.Timestamp created_ats_list = 4 [(tableau.field) = {name:"CreatedAts"}];
```
wenchy: > Auto-append "_list" suffix so users don't need to pluralize the name.

rename `CreatedAts` -> `CreatedAt`, as the protogen will auto add `_list` suffix

wenchy (changes requested):

wenchy (approved):

test/functest/proto/default/excel__struct__struct.proto
Kybxd: done

PR #450, opened 2026-08-13
fix(protogen): stop preserveFieldNumbers from polluting type infos

Kybxd:
## Problem

`preserveFieldNumbers` needs the previously generated protos parsed before `prepareOutdir` deletes them, but it reused the resulting registry as the type infos as well:

```go
protoRegistryFiles := gen.ProtoRegistryFiles
// preserveFieldNumbers also needs generated protos parsed before they
// are deleted below.
if useGeneratedProtos || gen.OutputOpt.PreserveFieldNumbers {
    protoRegistryFiles = gen.getProtoRegistryFilesWithGenerated()
}
gen.typeInfos = xproto.GetAllTypeInfo(protoRegistryFiles, gen.ProtoPackage)
```

As a result, the types only existing in the previously generated protos were mistaken for predefined ones.

### Failure path

With `GenAll` + `preserveFieldNumbers: true`, given a sheet referring to a type that was removed from the workbooks and now only exists in a previously generated proto:

1. the type is found in the polluted type infos, hence judged as `Predefined: true`;
2. the exporter emits `import "<that file>.proto"`;
3. but `GenAll` has already deleted that file and will never regenerate it.

So protogen succeeded and produced **uncompilable protos**. The expected behavior is a clear `predefined type not found` error.

That the type infos are meant to hold predefined types only is confirmed by them being populated dynamically during the first pass (`sheet_mode.go`, `protogen.go`), i.e. the baseline should contain nothing but predefined types.

## Fix

Decouple the cache warm-up from the type infos. `getProtoRegistryFilesWithGenerated` is guarded by `sync.Once`, so calling it merely to populate the cache is enough — the very same registry is still shared with `findMDFromGeneratedProtos` afterwards, keeping `preserveFieldNumbers` working, while the type infos are built from the predefined protos only, unless the advanced first-pass mode is enabled.

## Tests

Two regression tests, verified to be meaningful by temporarily reverting the fix:

| | before fix | after fix |
|---|---|---|
| `Test_preprocess_preserveFieldNumbersNotPollutingTypeInfos` | **FAIL** | PASS |
| `Test_preprocess_useGeneratedProtosPopulatesTypeInfos` | PASS | PASS |

The first one also asserts the warm-up still works, i.e. the previously generated proto is deleted by `prepareOutdir` yet still resolvable from the cached registry. The second one guards against overcorrecting: the advanced first-pass mode must keep seeing those types.

Full `go test ./...` passes.

### DIFF
--- a/internal/protogen/protogen.go
+++ b/internal/protogen/protogen.go
@@ -134,10 +134,19 @@ func (gen *Generator) preprocess(useGeneratedProtos, delExisted bool) error {
 	outdir := filepath.Join(gen.OutputDir, gen.OutputOpt.Subdir)
 	// parse custom imported proto files
 	protoRegistryFiles := gen.ProtoRegistryFiles
-	// preserveFieldNumbers also needs generated protos parsed before they
-	// are deleted below.
-	if useGeneratedProtos || gen.OutputOpt.PreserveFieldNumbers {
+	if useGeneratedProtos {
+		// NOTE: the advanced first-pass mode parses only the specified
+		// workbooks, so the previously generated protos are intended to serve
+		// as the predefined types of the others. They are not deleted in this
+		// mode, hence still importable.
 		protoRegistryFiles = gen.getProtoRegistryFilesWithGenerated()
+	} else if gen.OutputOpt.PreserveFieldNumbers {
+		// Cache the previously generated protos before they are deleted below,
+		// as they are not parsable any more once deleted. The result is
+		// deliberately discarded rather than used as the type infos, otherwise
+		// the types only existing in them would be mistaken for predefined
+		// ones, whose proto file is never regenerated in this run.
+		_ = gen.getProtoRegistryFilesWithGenerated()
 	}
 	gen.typeInfos = xproto.GetAllTypeInfo(protoRegistryFiles, gen.ProtoPackage)
 	return prepareOutdir(outdir, gen.InputOpt.ProtoFiles, delExisted)
--- a/internal/protogen/preprocess_test.go
+++ b/internal/protogen/preprocess_test.go
@@ -1,0 +1,93 @@
+package protogen
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/tableauio/tableau/internal/x/xfs"
+	"github.com/tableauio/tableau/options"
+)
+
+// prepareGeneratedProto writes a proto file into the output dir, mimicking a
+// previously generated one, whose message type is absent from both the input
+// workbooks and the predefined proto files.
+func prepareGeneratedProto(t *testing.T, outdir string) {
+	t.Helper()
+	require.NoError(t, os.MkdirAll(outdir, xfs.DefaultDirPerm))
+	content := `syntax = "proto3";
+package protoconf;
+option go_package = "github.com/tableauio/tableau/test/dev/protoconf";
+
+import "tableau/protobuf/tableau.proto";
+
+option (tableau.workbook) = {name:"Stale#*.csv" namerow:1 typerow:2 datarow:3};
+
+message StaleConf {
+  option (tableau.worksheet) = {name:"StaleConf"};
+
+  map<uint32, uint32> stale_map = 1 [(tableau.field) = {key:"ID" layout:LAYOUT_VERTICAL}];
+}
+`
+	require.NoError(t, os.WriteFile(filepath.Join(outdir, "stale_conf.proto"), []byte(content), xfs.DefaultFilePerm))
+}
+
+func newPreserveFieldNumbersGenerator(t *testing.T, subdir string) *Generator {
+	t.Helper()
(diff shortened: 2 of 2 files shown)


internal/protogen/protogen.go
```diff
@@ -132,11 +132,18 @@ func (gen *Generator) parseProtoRegistryFiles(useGeneratedProtos bool) (*protore
...
+	}
 	// parse custom imported proto files
 	protoRegistryFiles := gen.ProtoRegistryFiles
-	// preserveFieldNumbers also needs generated protos parsed before they
-	// are deleted below.
-	if useGeneratedProtos || gen.OutputOpt.PreserveFieldNumbers {
+	if useGeneratedProtos {
 		protoRegistryFiles = gen.getProtoRegistryFilesWithGenerated()
 	}
 	gen.typeInfos = xproto.GetAllTypeInfo(protoRegistryFiles, gen.ProtoPackage)
```
wenchy, line 151: But now, it still combines  `gen.getProtoRegistryFilesWithGenerated()` into gen.typeInfos when **useGeneratedProtos**.

internal/protogen/protogen.go
```diff
@@ -132,11 +132,18 @@ func (gen *Generator) parseProtoRegistryFiles(useGeneratedProtos bool) (*protore
...
 func (gen *Generator) preprocess(useGeneratedProtos, delExisted bool) error {
 	outdir := filepath.Join(gen.OutputDir, gen.OutputOpt.Subdir)
+	if gen.OutputOpt.PreserveFieldNumbers {
+		// Warm up the cache before the generated protos are deleted below, as
+		// they are not parsable any more once deleted.
+		//
+		// NOTE: the result is intentionally not used as the type infos below,
+		// otherwise the types only existing in the generated protos would be
+		// mistaken for predefined ones.
+		gen.getProtoRegistryFilesWithGenerated()
```
wenchy: Confusing: Return value not checked.

wenchy (changes requested):

internal/protogen/protogen.go
Kybxd: Good catch that the comment was misleading — it stated the result is "intentionally not used as the type infos" while the line right below did exactly that when `useGeneratedProtos` is true. Sorry for the confusion.

The behavior itself is intended though. `useGeneratedProtos=true` has exactly one callsite, `GenWorkbook` + `FirstPassModeAdvanced`, whose very definition is to parse based on the previously generated protos:

```go
FirstPassModeAdvanced FirstPassMode = "advanced" // parse based on all previous generated proto files
```

The asymmetry hinges on `delExisted`:

- `GenAll` (`delExisted=true`): the outdir is emptied, so treating a generated type as predefined makes the exporter emit an import of a file that is never regenerated — the bug this PR fixes.
- advanced (`delExisted=false`): nothing is deleted and only the specified workbooks are regenerated, so the generated protos of the others remain on disk. Importing them is correct and required, otherwise the mode would be pointless.

So I kept the logic and rewrote the comments to make the two cases explicit, side by side (also addressing your other comment about the discarded return value):

```go
protoRegistryFiles := gen.ProtoRegistryFiles
if useGeneratedProtos {
	// NOTE: the advanced first-pass mode parses only the specified
	// workbooks, so the previously generated protos are intended to serve
	// as the predefined types of the others. They are not deleted in this
	// mode, hence still importable.
	protoRegistryFiles = gen.getProtoRegistryFilesWithGenerated()
} else if gen.OutputOpt.PreserveFieldNumbers {
	// Cache the previously generated protos before they are deleted below,
	// as they are not parsable any more once deleted. The result is
	// deliberately discarded rather than used as the type infos, otherwise
	// the types only existing in them would be mistaken for predefined
	// ones, whose proto file is never regenerated in this run.
	_ = gen.getProtoRegistryFilesWithGenerated()
}
```

The `else if` also makes it clear the two are mutually exclusive, which the previous two separate `if`s did not. Verified equivalent for all four flag combinations (`sync.Once` makes the extra call in the old code a no-op).

internal/protogen/protogen.go
Kybxd: Agreed, and it was worse than it looked: `getProtoRegistryFilesWithGenerated` handles errors by `panic`, so that line silently hid both the discarded value and the fact that the error path panics.

Fixed by folding it into an `else if` next to the branch that does use the value, and marking the discard explicitly with `_ =` plus a comment stating why:

```go
} else if gen.OutputOpt.PreserveFieldNumbers {
	// Cache the previously generated protos before they are deleted below,
	// as they are not parsable any more once deleted. The result is
	// deliberately discarded rather than used as the type infos, otherwise
	// the types only existing in them would be mistaken for predefined
	// ones, whose proto file is never regenerated in this run.
	_ = gen.getProtoRegistryFilesWithGenerated()
}
```

Happy to go further and change `getProtoRegistryFilesWithGenerated` to return an `error` instead of panicking, so this becomes a checked call — but that also touches `findMDFromGeneratedProtos`, so I left it out to keep this PR focused on the bug fix. Let me know if you'd prefer it here.

wenchy (approved):

